PR #229, opened 2021-09-16
Enabling mypy type-checking

marcromeyn:


merlin_standard_lib/utils/proto_utils.py
```diff
@@ -51,7 +51,7 @@ def proto_text_to_better_proto(
     proto_text = path_proto_text
     if os.path.isfile(proto_text):
         with open(path_proto_text, "rb") as f:
-            proto_text = f.read()
+            proto_text = str(f.read())
```
benfred, line 54: This is incorrect as written - and will produce garbage output (fwiw, for converting from bytes to str values is done with something like ```f.read().decode("utf8")```

This code will return the str repr of the bytes object (including the "b" marker). Like running this code

```python
filename = "test.txt"
with open(filename, "w") as f:
    f.write("test")
with open(filename, "rb") as f:
    text = str(f.read())
    print(f"The text is '{text}'")
```
will print out ```The text is 'b'test''``` 

Instead of this how about doing just open the file as text? 
```python
        with open(path_proto_text, "r") as f:
            proto_text = f.read()
```

marcromeyn: Ah good catch!

PR #245, opened 2021-09-22
Add LabelSmoothCrossEntropyLoss util class

### LINKED ISSUE #104: Add utils for  `LabelSmoothCrossEntropyLoss`
- Add a utils for LabelSmoothCrossEntropyLoss defined at :  https://github.com/NVIDIA-Merlin/Transformers4Rec/blob/ea1f2ff46520026e0b293e07384dc535a20faeee/transformers4rec/recsys_meta_model.py#L1198 

- The [Label smoothing](https://arxiv.org/abs/1906.02629) consists of transforming class labels into soft targets so as to reduce the model's over confidence.

sararb:
1. Please write a description in this text box of the changes that are being
   made.

This PR adds `LabelSmoothCrossEntropyLoss` to PyTorch examples_utils to allow using label smoothing technique in NextItemPredictionTask. 

2. Please ensure that you have written units tests for the changes made/features
   added.

A unit test is added to check the equality between LabelSmoothCrossEntropyLoss set in the ItemPredictionTask and the manual calculation of smoothed cross-entropy loss. 

3. If you are closing an issue please use one of the automatic closing words as
   noted here: https://help.github.com/articles/closing-issues-using-keywords/

closes #104


tests/torch/utils/test_examples_utils.py
```diff
@@ -0,0 +1,39 @@
+import pytest
+
+pytorch = pytest.importorskip("torch")
+tr = pytest.importorskip("transformers4rec.torch")
+examples_utils = pytest.importorskip("transformers4rec.torch.utils.examples_utils")
+
+
+@pytest.mark.parametrize("label_smoothing", [0.1, 0.6])
```
gabrielspmoreira, line 8: You could include `label_smoothing=0.0` as a test case here, just to ensure it does not fail with the default 0 value

transformers4rec/torch/utils/examples_utils.py
```diff
@@ -90,3 +92,62 @@ def fit_and_evaluate(trainer, start_time_index, end_time_index, input_dir):
         trainer.wipe_memory()
 
     return aot_metrics
+
+
+class LabelSmoothCrossEntropyLoss(_WeightedLoss):
```
gabrielspmoreira, line 97: What if we move `LabelSmoothCrossEntropyLoss` from `examples_utils.py` to a Python file deticated to losses. It will make it easier to find the documentation under e.g. 'Transformers4rec.torch.losses' and we can add the upcoming pairwise losses their too

gabrielspmoreira (approved):
Just left some minor suggestions. But I am ok with this PR.

transformers4rec/torch/utils/examples_utils.py
sararb: you're right ! I was not sure about whether we want to expose the losses module in this early version or not. I have just moved it to losses.py module.

PR #247, opened 2021-09-22
Refactoring github workflows + adding Makefile

marcromeyn:
### Goals :soccer:

- Refactoring github workflows to run tests + linting separately.
- Introducing docs deploy.
- Simplifying running tests + docs locally by introducing a Makefile


.github/workflows/docs.yml
```diff
@@ -0,0 +1,19 @@
+name: docs
```
benfred, line 1: Can you test out this workflow on a fork first?  I'd like to see what the docs being pushed out by this workflow look like before merging

I'm also not convinced this is quite right - we're pushing to a 'main' subfolder in gh-pages (to match eventual sphinx-multiversion support), and I think we might need dependencies like TF/PT installed first to generate the API documentation.

README.md
```diff
@@ -1,6 +1,6 @@
 # [Transformers4Rec](https://github.com/NVIDIA-Merlin/Transformers4Rec/) | [Documentation](https://nvidia-merlin.github.io/Transformers4Rec/main/)
 <p align="left">
-    <a href="https://github.com/NVIDIA-Merlin/Transformers4Rec/actions/workflows/ci.yml">
+    <a href="https://github.com/NVIDIA-Merlin/Transformers4Rec/actions/workflows/tests.yml">
         <img alt="Build" src="https://img.shields.io/github/checks-status/NVIDIA-Merlin/Transformers4Rec/main">
```
benfred, line 4: The image at this url just says 'Checks: Pending'  - can we use the github actions one instead? https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/adding-a-workflow-status-badge

Like https://github.com/NVIDIA-Merlin/Transformers4Rec/actions/workflows/ci.yml/badge.svg?branch=main currently - or https://github.com/NVIDIA-Merlin/Transformers4Rec/actions/workflows/tests.yml/badge.svg?branch=main after this change ?

docs/source/api/transformers4rec.data.rst
```diff
@@ -0,0 +1,53 @@
+transformers4rec.data package
```
benfred, line 1: I don't think we want to include these package yet (because of that betterproto issue during docs build)

docs/source/api/modules.rst
```diff
@@ -1,8 +1,7 @@
-API Documentation
-=================
+transformers4rec
```
benfred, line 1: I know this is autogenerated - but we should have the previous 'API Documentation ' header here.

jperez999:
rerun tests

benfred:
I don't think this PR is necessary anymore - we have docs being built and pushed out automatically now https://github.com/NVIDIA-Merlin/Transformers4Rec/blob/main/.github/workflows/docs-sched-rebuild.yaml

PR #302, opened 2021-11-01
Add to_tf_model to T4RecConfig class 

sararb:
### Goals :soccer:
The TensorFlow [example](https://github.com/NVIDIA-Merlin/Transformers4Rec#quick-tour) in the read.me uses the method `to_tf_model` to ensure compatibility with torch API but it is not implemented.  This PR adds this method to T4RecConfig.

### Testing Details :mag:
Add a unit test of the `to_tf_model` method to test_model.py that tests the conversion of all supported HF transformers config to a `Model` class and checks its output shape.

### DIFF
--- a/tests/tf/model/test_model.py
+++ b/tests/tf/model/test_model.py
@@ -16,6 +16,8 @@
 
 import pytest
 
+from transformers4rec.config import transformer as tconf
+
 tf = pytest.importorskip("tensorflow")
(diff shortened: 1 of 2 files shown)


marcromeyn (approved):

gabrielspmoreira (approved):
Sounds good to me

PR #308, opened 2021-11-02
Refactor RankingMetric to fix serialization and graph-mode

### LINKED ISSUE #283: [BUG] Add tf graph-mode support

1.  Define the model for NextItemPrediction
2. Compile model with graph model: `model.compile(optimizer="adam", run_eagerly=False)
More details can be found  in this unit [test](https://github.com/NVIDIA-Merlin/Transformers4Rec/blob/7aca855103c2243f32f69c7a6452ea67320406b8/tests/tf/model/test_model.py#L107) 

### Expected behavior

Making sure the tf pipeline (training, evaluating, and saving the model) is working in `graph mode`

sararb:
Fixes #283 

### Goals :soccer:

- [x] Re-factor the tf ops used to define the RankingMetric classes to make sure they are working in both modes: `eager` and `graph`. 

- [x] Define `get_config` / `from_config` methods to the custom classes for serialization.

- [x] Add a non trainable tf.Variable to `RankingMetric` classes as placeholder to store the updated states at each iteration. This is needed in `graph` mode as the metrics are part of the Keras class `NextItemPredictionTask`.

- [x]   The method `process_metric` is added to flatten the list of scores returned by ranking metrics with respect to their threshold

### Testing Details :mag:

- Add `test_score_different_from_zero`  to check that computed metrics are different from 0.  
- Add `test_serialization_ranking_metric` to check the serialization of RankingMetric classes.

### DIFF
--- a/transformers4rec/tf/ranking_metric.py
+++ b/transformers4rec/tf/ranking_metric.py
@@ -78,12 +78,12 @@ class RankingMetric(tf.keras.metrics.Metric):
         if self.labels_onehot:
             y_true = tf_utils.tranform_label_to_onehot(y_true, tf.shape(y_pred)[-1])
 
-        metric = self._metric(
+        self._metric(
             ks=tf.convert_to_tensor(self.top_ks),
             scores=tf.reshape(y_pred, [-1, tf.shape(y_pred)[-1]]),
             labels=y_true,
         )
-        self.metric_mean.append(metric)
+        self.metric_mean.append(self.accumulator)
 
     def result(self):
         # Computing the mean of the batch metrics (for each cut-off at topk)
@@ -93,7 +93,9 @@ class RankingMetric(tf.keras.metrics.Metric):
         self.metric_mean = []
 
     @abstractmethod
-    def _metric(self, ks: tf.Tensor, scores: tf.Tensor, labels: tf.Tensor) -> tf.Tensor:
+    def _metric(
+        self, ks: List[int], scores: tf.Tensor, labels: tf.Tensor, return_value: bool = False
+    ) -> tf.Tensor:
         """
         Compute ranking metric over predictions and one-hot targets for different cut-offs.
         This method should be overridden by subclasses.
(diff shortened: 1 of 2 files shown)


transformers4rec/tf/ranking_metric.py
```diff
@@ -84,54 +105,66 @@ def _metric(self, ks: tf.Tensor, preds: tf.Tensor, target: tf.Tensor) -> tf.Tens
...
+            )
 
-        for index, k in enumerate(ks):
-            precisions[:, index].assign(tf.reduce_sum(topk_labels[:, : int(k)], axis=1) / float(k))
+            self.accumulator.scatter_nd_update(
+                indices=indices, updates=tf.reduce_sum(topk_labels[:, : int(k)], axis=1) / float(k)
+            )
 
-        return precisions
+        return self.accumulator
```
marcromeyn, line 140: Might be a little cleaner to have this method not return anything and just update `self.accumulator`. In `update_state` from the base-class we can just pass this variable on.

marcromeyn (approved):
Looks good, just added a small nit-pick.

transformers4rec/tf/ranking_metric.py
sararb: yeah your right, I return it in the `_metric` because we are using this method as a nested call in `_metric` of AvgPrecisionAt  [here](https://github.com/NVIDIA-Merlin/Transformers4Rec/blob/c9aae3fedd1b1f71be6d93da3f4056808b7e27a4/transformers4rec/tf/ranking_metric.py#L220)  
Maybe we can have a flag in `_metric` to decide whether we want to return the tensor or not ?

marcromeyn: So the reason I brought this up was that implementing the `_metric` function is potentially user-code, since we want to allow users to create custom metrics. It would therefore be useful to make implementing this function as easy as possible. It's a bit tricky now that as a user now you have to know to update self.accumulator + returning it. Therefore I thought it would be simpler to just update the self.accumulator and we take care of the rest. We could create a public method that can be called if you want to call the _metric function and get the output back. What do you think?

sararb: Thank you for the explanation! I totally agree with you. Regarding the name of the public method, can we call it something like `metric_fn`?

sararb: I added the public method `metric_fn` for the cases where we need the output back and updated `_metric`.

marcromeyn (approved):

PR #311, opened 2021-11-03
Disable stochastic swap noise during eval

### LINKED ISSUE #310: [BUG] Stochastic Swap Noise is being applied during eval on PyT API
### Bug description
The Stochastic Swap Noise transformation should be applied only during training, but is being also applied during evaluation in the PyTorch API

Put the `StochasticSwapNoise` in the eval mode and the random replacement of items will be applied

### Additional context
This issue does not occur with the TF API, where swap noise is correctly applied during training

gabrielspmoreira:
Fixes #310 

### Goals :soccer:
- Fixes Stochastic swap noise, which was being applied during eval in the PyT API

- Now it checks if the module is in the `eval` mode before applying the transformation

### Testing Details :mag:
- Added tests to PyT and TF apis to check for regression on this issue


marcromeyn (approved):

sararb (approved):
Looks good to me!

PR #315, opened 2021-11-05
Adding missing DLLogger requirement for the paper reproducibility example

gabrielspmoreira:
### Goals :soccer:
- Adds the missing NVIDIA DLLogger requirement for the paper reproducibility demo


marcromeyn:
Wouldn't this be better as an optional dependency?

benfred (approved):

benfred:
> Wouldn't this be better as an optional dependency?

Isn't this requirements.txt file already optional? (like it'll only be installed if you go ```pip install -r examples/t4rec_paper_experiments/requirements.txt``` and isn't run as part of the main transformers4rec install).

It also doesn't look like dllogger is optional to run the scripts here : https://github.com/NVIDIA-Merlin/Transformers4Rec/blob/5b7e9be2d98e03c9c8808e79eff48ddf9e807ed6/examples/t4rec_paper_experiments/t4r_paper_repro/exp_outputs.py#L21-L23

gabrielspmoreira:
Indeed @benfred. This requirement is only necessary to run the paper reproducibility example, that is why it is in a separate `requirements.txt`. To make this dependency optional, we would have to change the code to only log using DLLogger when it is installed. 
P.s. For our internal experiments at NVIDIA, DLLogger is always required because it saves the final metrics in the expected format by AutoBench, our manager of distributed jobs. DLLogger it is an [open-source project ](https://github.com/NVIDIA/dllogger) and is also required by NVIDIA Deep Learning examples

PR #316, opened 2021-11-05
Add codespell to CI

benfred:
Adds codespell to our CI process and precommit hooks. This catches common spelling mistakes before
they are merged


gabrielspmoreira (approved):
That is a good addition! It seems we don't have that many spelling errors in this repo :)

PR #317, opened 2021-11-05
Fix save/load tf4rec model

### LINKED ISSUE #313: [BUG] Loading the SavedModel is re-building the graph in the wrong order.
### Bug description

- Loading the custom class `tr4rec.tf.Model` from saved config  raises the following issue: 
```    
@property
def item_embedding_table(self):
        assert self.item_id is not None
    
>       return self.embedding_tables[self.item_id]
E       AttributeError: 'SequenceEmbeddingFeatures' object has no attribute 'embedding_tables'
```
  ==> Trying to build the head before the body is raising this error as the `NextItemPredictionTask` is requiring the body to be built to retrieve the item embedding table.  

- This is related to the fact that Keras `.load` method (definition [here](https://github.com/keras-team/keras/blob/3a33d53ea4aca312c5ad650b4883d9bac608a32e/keras/saving/saved_model/load.py#L79)) build the graph from the SavedModel in the wrong order (head -> task -> body) instead of (body -> head -> task )

- More details when trying to debug the error: 
     - As we are decoupling the call to `body` object from the call of the `head` in the Model `call` method (see code [here](https://github.com/NVIDIA-Merlin/Transformers4Rec/blob/ff0e5910142af91b8bc5b619d7b4fd753a36649b/transformers4rec/tf/model/base.py#L512#L513)),  the saved graph is not linking the `body` to `Head` class, and we get the  parent/children structures:  head -> Task -> body 

### Steps/Code to reproduce the bug
1.  save the NextItemPrediction model in the Read.me file using :   `model.save(path_to_dir)`
2. Loading it back using `model = tf.keras.models.load_model(path_to_dir,  compile=True)`

### Expected behavior

- When loading the custom class T4Rec `Model` defined here,  the [KerasObjectLoader](https://github.com/keras-team/keras/blob/3a33d53ea4aca312c5ad650b4883d9bac608a32e/keras/saving/saved_model/load.py#L244) should build the graph in the following order:  body -> head -> task.  

### Environment details
- Transformers4Rec version: The code from the PR #308 
- Python version: 3.8
- Huggingface Transformers version: 4.12.*
- Tensorflow version (GPU?): No

### Additional context

Some investigated ideas:

1.  Load the graph without building it from config: Keras .load() does not offer this option, and the automatic building from config is done by default

2. Specify the order of the building execution: Keras .load() automatically infers the order from the `node ids` and [dependencies](https://github.com/keras-team/keras/blob/3a33d53ea4aca312c5ad650b4883d9bac608a32e/keras/saving/saved_model/load.py#L310) created when saving the model metadata. This metadata is generated based on the `call` methods of all nested layers used in `tr.Model.` 

3.  Always call the `body` within the head `call` method. The issue raises for NextItemPredictionTask as it requires the `body` when building the layers [here](https://github.com/NVIDIA-Merlin/Transformers4Rec/blob/ff0e5910142af91b8bc5b619d7b4fd753a36649b/transformers4rec/tf/model/base.py#L369), which is happening before the call (At this moment, the body is not built, and we cannot access the item embedding table)

4.  (Not tested yet): Refactor Item Prediction task to get the item-embedding table outside the `build` method? This could be done via the definition of a private class method `_build` that can be called within the `call` of the task when `weight_tying` is True and `self.item_embedding_table` is not set.

sararb:
Fixes #313 

### Goals :soccer:
- This PR fixes the saving and loading  T4Rec `Model` classes.

1. Register the custom class Model within Translofmers4Rec; this allows us to call the `load` method without specifying custom objects.

2. Expose `compute_metrics` flag in compute_loss method of `Model` class to compute metrics during train / evaluation.

3. Refactor the build method of the `Head` class to control what modules to build first. 

4. Set the `call` of the `Head` class to call the `body` by default. 

5. Add build method to `MergeTabular`  to build the `EmbeddingFeatures` modules outside the call method. This is needed when building the `NextItemPredictionHead` inside the `Head` class as it requires access to the item-embedding weights defined within the `SequenceEmbeddingFeatures` module. 

6. Add `embedding_tables` as a part of the `EmbeddingFeatures` class to load them from saved checkpoints. Note that during the build of the class, we only add the weights of new features. 

### Testing Details :mag:
Refactor test_model to remove redundant code and add a unit test related to saving/loading Transformer-based models.


albert17:
rerun tests

tests/tf/model/test_model.py
```diff
@@ -85,83 +85,65 @@ def test_to_tf_model_from_config(yoochoose_schema, tf_yoochoose_like, config_cls
...
+):
+    model = tf_next_item_prediction_model(config=config)
+    model.compile(optimizer="adam", run_eagerly=run_eagerly)
+
+    dataset = tf.data.Dataset.from_tensor_slices(
+        (tf_yoochoose_like, tf_yoochoose_like["item_id/list"])
+    ).batch(50)
+
+    inputs = next(iter(dataset))[0]
+    model._set_inputs(inputs)
```
gabrielspmoreira, line 143: Why are we calling _set_inputs() method here?

gabrielspmoreira (approved):

tests/tf/model/test_model.py
sararb: To save the model, we need to set the input shapes of all sub-modules, and this is automatically done when we call the `fit` or `call` methods of the model. But here, we only focus on the saving part, so we needed to explicitly call `model._set_inputs(...)` to set the shapes.

PR #326, opened 2021-11-15
CI working

albert17:
Fixes spaces for building tf4rec

PR #331, opened 2021-11-16
Fix GPU CI

benfred:
The gpu CI runner we have set up runs the 'build_and_test.sh' script - which wasn't
working. Fix


benfred:
duplicate of #326

PR #332, opened 2021-11-18
Getting-started notebook with CLM task 

sararb:
### Goals :soccer:
- The PR introduces the first example of getting-started notebook using Tensorflow API. 

-  Add the first version of the getting-started notebook for TensorFlow, similar to the one we have in pytorch API.
- Some issues were raised when defining the notebook: 

     - `NDCG` metric was returning the same score for all thresholds: This was related to the fact that `gains` and `normalized_gains` are pointing to the same variable `self.accumulator` of DCGAt metric and therefore taking the same value. It was fixed by using the self.accumulator of the NDCGAt instead of gains when computing the updates. 
   
     -  Update `process_metrics` to print the ranking metrics in the same format we used for pytorch API. 
     
     - `mlm` is leading to evaluation scores close to 1 and the issue (#333 ) will be tracked in a different PR. 

### Testing Details :mag:
- Add a unit test that checks if the ranking scores of different thresholds are not equal.

### DIFF
--- a/examples/getting-started-session-based/03-session-based-XLNet-with-TF.ipynb
+++ b/examples/getting-started-session-based/03-session-based-XLNet-with-TF.ipynb
@@ -2,8 +2,8 @@
  "cells": [
   {
    "cell_type": "code",
-   "execution_count": 1,
-   "id": "882fa1cb",
+   "execution_count": null,
+   "id": "470172f1",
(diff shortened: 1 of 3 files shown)


albert17:
rerun tests

rnyak (approved):

Issue #328, opened 2021-11-15
[FEA] Create automated Integration & Performance Tests for NVTabular + Transformers4Rec pipeline


gabrielspmoreira:
# 🚀 Feature request

- [ ] Add integration tests of a pipeline with NVTabular + Transformers4Rec library, based on a few experiments inspired in the RecSys'21 paper. I
- [ ] Log accuracy and performance (speed) metrics and fail if they change drastically

## Motivation

- Regression in accuracy and performance can happen both due to changes in NVTabular, Transformers4Rec or the external HuggingFace Transformers and we need to track that.

gabrielspmoreira:
@albert17 here are the integration tests scripts of Transformers4Rec library using the PyTorch API and the paper reproducibility script. It is a total of 7 tests, which test different Transformer architectures and input feature sets with the same hyperparameters from the best runs reported in the paper. The script take as input a [small version of the REES46 dataset](https://drive.google.com/file/d/1payLwuwfa_QG6GvFVg4KT1w7dSkO-jPZ/view?usp=sharing) (only two days) and uses the NVTabular Dataloader for PyTorch to load those parquet files.

## Container
The integration tests should run on the latest Merlin PyTorch Training nightly container. They can also run on the latest stable container `nvcr.io/nvidia/merlin/merlin-pytorch-training:21.11`, if the latest code is pulled in the container folders `/nvtabular` and `/transformers4rec`.

### Setup

#### Installing dependencies
```bash
# To fix the PyTorch error with DLFW pytorch: RuntimeError: unique_by_key: failed on 2nd step: cudaErrorInvalidValue: invalid argument
pip install torch==1.10
# Installing the paper repro script specific requirements
pip install -r /transformers4rec/examples/t4rec_paper_experiments/requirements.txt
```

#### Downloading and unzipping he REES46 small dataset
```bash
pip install gdown
gdown https://drive.google.com/uc?id=1payLwuwfa_QG6GvFVg4KT1w7dSkO-jPZ
DATASET_PATH=rees46/
unzip -d $DATASET_PATH "rees46_ecom_dataset_small_for_ci.zip"
```

#### Integration tests
[...]

Issue #340, opened 2021-12-01
[FEA] Unit tests can be run as a single command


albert17:
Currently, if we run tests as `pytest /transformers4rec/tests` all the tests are skipped in tf and pytorch container because of pyt/tf packages are missing respectively.

More details to be added

PR #343, opened 2021-12-05
Integration testing infrastructure

albert17:
Addresses #328


albert17:
@gabrielspmoreira I am getting this error when trying to run the tests 
```
/usr/bin/python3: Error while finding module specification for 't4r_paper_repro.transf_exp_main' (ModuleNotFoundError: No module named 't4r_paper_repro')
```

albert17:
@gabrielspmoreira I have been able to get it running by calling directly the script, as it is not installed as a module.

Getting an OOM error when using a V100 16GB GPU
```
RuntimeError: CUDA out of memory. Tried to allocate 2.43 GiB (GPU 0; 15.77 GiB total capacity; 4.72 GiB already allocated; 326.25 MiB free; 4.73 GiB reserved in total by PyTorch) If reserved memory is >> allocated memory try setting max_split_size_mb to avoid fragmentation.  See documentation for Memory Management and PYTORCH_CUDA_ALLOC_CONF
12/05/2021 19:22:04 - INFO - numba.cuda.cudadrv.driver -   add pending dealloc: module_unload ? bytes
```

gabrielspmoreira:
@albert17 I have tested those training arguments in a V100 32 GB. Can you pls try using 128 a smaller batch size (--per_device_train_batch_size 128 --per_device_eval_batch_size 128) in the commands?

gabrielspmoreira:
@albert17 Should we run all integration tests in a single bash script? With "set -e" in the begining of the script it seems it will stop running if any command fails. Ideally tests should run independently of each other, so that if one test fails others can run.
Would that be possible?

albert17:
@gabrielspmoreira changing batch to 128 works. I also removed the `set -e` flag

ci/test_unit.sh
```diff
@@ -0,0 +1,18 @@
...
+
+## Tensorflow container
+if [ "$container" == "merlin-tensorflow-training" ]; then
+    pytest tests/tf
+# Pytorch container
+elif [ "$container" == "merlin-pytorch-training" ]; then
+    pytest tests/torch
+# Inference container
+elif [ "$container" == "merlin-inference" ]; then
+    pytest tests
```
benfred, line 17: Do we need to differentiate between the containers on running the unittests? 

The transformers4rec tests here do things like ```torch = pytest.importorskip("torch")``` to skip if we don't have torch or tensorflow installed. (also I don't think the inference container has TF - so we're already relying on this behaviour?).

benfred (approved):
lgtm

ci/test_unit.sh
albert17: @benfred We already have these kind of checks
For TF: https://github.com/NVIDIA-Merlin/Transformers4Rec/blob/main/tests/torch/conftest.py#L19
For PyT: https://github.com/NVIDIA-Merlin/Transformers4Rec/blob/main/tests/tf/conftest.py

However, running `pytest tests` and reaching those skipping imports makes the whole testing being skipped. I don't know why. May it be because of subfolders structure? I created an issue in the past: https://github.com/NVIDIA-Merlin/Transformers4Rec/issues/340

This is why I created the script specifying containers

benfred: thanks for the clarification!

gabrielspmoreira:
Looks good to me @albert17 !

PR #381, opened 2022-03-15
fix: Pin mypy version to fix CI

mikemckiernan:
I have a docs PR that I want to contribute, but some testing issues are clouding my view.

I took a guess at a fix.


mikemckiernan:
Someone else fixed this.

PR #436, opened 2022-06-20
Add the support of prediction step to the Trainer class

sararb:
### Goals :soccer:

- The objective of this PR is to add the support of generating the predictions during the `Test` in the `Trainer` class. This required two fixes: 
- [x]  Setting the test dataloader  
- [x] Add an `ignore-masking` argument to leverage the whole input sequence during the Test prediction step. 

- Add the `ignore_masking` option to the forward method of `TabularSequenceFeatures` to check whether masking should be applied or not. 
- Update the Trainer `prediction_step` loop to set the `ignore_masking` based on the current step description (Prediction or  not)
- Add `get_test_dataloader` method to the `Trainer`.

### Testing Details :mag:
- Extend the existing test of the Trainer to check the prediction-step
- Add a unit test of the `TabularSequenceFeatures` module to make sure `ignore-masking` is correctly set.


transformers4rec/torch/model/prediction_task.py
```diff
@@ -218,16 +218,22 @@ def forward(self, inputs: torch.Tensor, **kwargs):
...
             labels = self.embeddings.item_seq
-
-        # remove padded items
-        trg_flat = labels.flatten()
-        non_pad_mask = trg_flat != self.padding_idx
-        labels_all = torch.masked_select(trg_flat, non_pad_mask)
-        x = self.remove_pad_3d(x, non_pad_mask)
+            non_pad_mask = labels != self.padding_idx
+            last_item_sessions = non_pad_mask.sum(dim=1) - 1
+            rows_ids = torch.arange(labels.size(0), dtype=torch.long, device=labels.device)
```
gabrielspmoreira, line 233: Doesn't this change the behavior compared to what we had before? I understand that before when `self.masking == False` the full sequence of items were used as label. Was `self.masking == False` only for predict? Or do we have other use cases where masking should be False to take into account?

For `model.predict()` I think it makes sense to mask only the last item, assuming it is a dummy item added in the end of the sequence when using MLM just to enforce predicting the last item using the whole sequence.

sararb: From our discussion yesterday, we should always set masking (`self.masking != None`) for the NextItemPredictionTask as the current implementation does not support explicit targets.  So I added a required check in the build method to make sure masking is present in the input Block. 

For the predict() step, we set the flag `ignore_masking=True` so that the whole sequence is considered and there is no need to add the dummy item in the end. The final prediction is computed based on the hidden states of the last position in the sequence (as the weights are shared between positions)

gabrielspmoreira (approved):

PR #444, opened 2022-06-27
Add PostContextFusion block to support Latent Cross technique

sararb:
### Goals :soccer:

- Add the [Latent Cross](https://static.googleusercontent.com/media/research.google.com/en//pubs/archive/46488.pdf) technique as a PostFusionContext block that merges the output of the sequential module with the contextual vector. This architecture was used in [SIGIR'21](https://arxiv.org/pdf/2107.05124.pdf) and Recsys'22 solutions. 

- Add `PostContextFusion` block to an experimental file.
-  It supports different fusion aggregation types

### Testing Details :mag:
- Add unit test for the `PostContextFusion`


gabrielspmoreira (commented):
I understand that we feed inputs to both `sequential_module` and `post_context_module` and we assume that the `post_context_module` is going to combine all context features in a single vector to combine with the output of `sequential_module` right?

I am wondering if this approach would cover more advanced use cases like our solution for the SIGIR, where the post-fusion context was composed by:
1. A non-sequential feature (e.g. the event type feature of the target)
2. The average of embeddings from a sequence different from the one provided as input to `sequential_module` (search events)

I think it is fine if for the first version we don't support both types of contextual info at the same time for post-fusion, but if we do we could maybe create a test demonstrating that

gabrielspmoreira (approved):

karunaahuja:
I am wondering if this approach would cover more advanced use cases like our solution for the SIGIR, where the post-fusion context was composed by:

A non-sequential feature (e.g. the event type feature of the target)
The average of embeddings from a sequence different from the one provided as input to sequential_module (search events)

> Following up on this. From the code it looks like the post_context is a sequence feature. How do we use other context feature: eg - user_embeddings with this module

PR #449, opened 2022-07-07
Set `INPUT_DATA_DIR` env var to `/tmp/data` in notebook tests

karlhigley:
Otherwise, the notebook tests fail in Jenkins, where the `jenkins` user doesn't have access to `/workspace/data`.

Fixes # (issue)

PR #450, opened 2022-07-07
Update requirement on `nvidia-dllogger` to follow install instructions

karlhigley:
This package isn't available from PyPI, which seems to be where our integration tests are looking for it. The README [suggests installing it](https://github.com/NVIDIA/dllogger#installation) this way.


karlhigley:
Blocked by #449

gabrielspmoreira (approved):

PR #454, opened 2022-07-11
Change the metric names prefix to align with the HF trainer code.

sararb:
### Goals :soccer:

HuggingFace [trainer](https://github.com/huggingface/transformers/blob/ac98a88fbc6377f93e8b7fbd244b0c3331bb82a0/src/transformers/trainer.py#L2924) class is expecting metrics prefixed with the following pattern `{metric_key_prefix}_`. In T4Rec, we were using the `\` separator to append the prefix to the metric name while HF is using the `_` separator. This PR is a quick fix that replaces `\` with `_` to align with HF code. 

### Testing Details :mag:
Updates the related tests that check the output metric names.

### DIFF
--- a/tests/torch/test_trainer.py
+++ b/tests/torch/test_trainer.py
@@ -176,15 +176,15 @@ def test_trainer_eval_loop(torch_yoochoose_next_item_prediction_model):
 
     assert isinstance(eval_metrics, dict)
     default_metric = [
-        "eval/next-item/ndcg_at_10",
-        "eval/next-item/ndcg_at_20",
-        "eval/next-item/avg_precision_at_10",
-        "eval/next-item/avg_precision_at_20",
-        "eval/next-item/recall_at_10",
-        "eval/next-item/recall_at_20",
+        "eval_next-item/ndcg_at_10",
+        "eval_next-item/ndcg_at_20",
+        "eval_next-item/avg_precision_at_10",
+        "eval_next-item/avg_precision_at_20",
+        "eval_next-item/recall_at_10",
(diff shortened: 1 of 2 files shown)


gabrielspmoreira:
@sararb . I think we were using the "/" in metric names so that they are better organized (by section) in logging visualization tools like TensorBoard and Weight&Biases.
What is the implication on W&B Trainer if not using their convention?

sararb:
> @sararb . I think we were using the "/" in metric names so that they are better organized (by section) in logging visualization tools like TensorBoard and Weight&Biases. What is the implication on W&B Trainer if not using their convention?

Thanks, Gabriel! So I come across that issue when trying to save the model's checkpoints for only the best epoch (based on a specified metric).  You can check [here](https://github.com/huggingface/transformers/blob/a462fc9232eb9e04f9bc9c710d8c9a3ac21056de/src/transformers/trainer.py#L2038) that there is a check about whether the metric is starting with 'eval_' and if not the prefix is added. This is raising a key error in T4Rec because the new name would be `eval_eval/{metric_name}` which is not the one returned by the model.  I will also update the PR with a unit test related to this specific use-case.

sararb:
> As suggested in the meeting, having eval metrics prefixed by `eval_/` might solve the issue?

Yes it could solve the issue, the user should just specify the best metric name with '/' at the begining: something like that `metric_for_best_model="/next-item/recall_at_10"`

I updated the code with the new separator '_/' and added a unit test for saving checkpoints based on best metric.

gabrielspmoreira (approved):

PR #483, opened 2022-09-01
[BUG] trainer.model.module renamed and DataParallel mode fixed

### LINKED ISSUE #473: [BUG] DataParallel training with Trainer is not using multiple GPUs
### Bug description
Before, when multiple GPUs were available (e.g. `CUDA_VISIBLE_DEVICES=0,1`) the model was automatically [wrapped](https://github.com/huggingface/transformers/blob/d6eeb871706db0d64ab9ffd79f9545d95286b536/src/transformers/trainer.py#L1327) by `DataParallel()` by HuggingFace Trainer class, which we inherit in our PyT API.
But it seems that only a single GPU is being used in such cases

1. Open the (02-End-to-end-session-based-with-Yoochoose-PyT.ipynb)[https://github.com/NVIDIA-Merlin/Transformers4Rec/blob/main/examples/end-to-end-session-based/02-End-to-end-session-based-with-Yoochoose-PyT.ipynb] notebook
2. Set `CUDA_VISIBLE_DEVICES=0,1` env variable
3. Train the model with `trainer.train()`
4. You will observe that only a single GPU is being used during training

### Expected behavior
Multiple GPUs should be used

### Environment details
- Transformers4Rec version: 4.12
- Merlin image: Merlin PyTorch 22.07

nzarif:
Fixed #473
`status/needs-review`

### Goals ⚽

Trainer was using only 1 GPU when DataParallel mode was set, fixed that so now training will be done on all the available GPUs when using DataParallel mode.

### Implementation Details 🚧

The `Trainer` wraps the input model using `class HFWrapper` which has an attribute named `module`. This overrides another attribute from [HF transformers Trainer](https://github.com/huggingface/transformers/blob/main/src/transformers/trainer.py) class with the same name. The overridden attribute prevents HF-transofrmer's `Trainer` class from wrapping the input model in `torch.nn.DataParallel` within its [`_wrap_model()` function](https://github.com/huggingface/transformers/blob/142e12afb457e31ec0e7323d9eba7b8bfee46c57/src/transformers/trainer.py#L1314). Renaming the `HFWrapper` attribute to something other than `module` solves the problem.

### Testing Details 🔍

Use any test including `Trainer.train()` and follow this steps:

1. Set `CUDA_VISIBLE_DEVICES=0,1` env variable
2. Train the model with `trainer.train()`
3. use `nvidia-smi` to monitor both GPUs being used during training.


gabrielspmoreira (approved):
Awesome @nzarif . Thanks for the fix!

It worth highlighting that you have tested this fix with our integration tests, which uses the script for paper reproducibility, and they started using multiple GPU when available. You mentioned me that only TransformerXL (`transfoxl`) raised an error in `DataParallel` mode, which is a know issue we had identified in the past (May 2021) during the paper experiments, as documented in the Notes [here](https://github.com/NVIDIA-Merlin/Transformers4Rec/tree/recsys2021/hf4rec)

PR #486, opened 2022-09-08
Adjust the device used in synthetic data generation

karlhigley:
It's important to use data on the same device that the model will be on during serving when tracing the model, so this makes it possible to generate synthetic data on GPU.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,25 +1,25 @@
 repos:
       - repo: https://github.com/timothycrosley/isort
-        rev: 5.6.4
+        rev: 5.10.1
         hooks:
         - id: isort
           additional_dependencies: [toml]
       - repo: https://github.com/python/black
-        rev: 20.8b1
+        rev: 22.8.0
         hooks:
         - id: black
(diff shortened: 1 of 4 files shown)


karlhigley:
rerun tests

oliverholworthy (approved):
Looks like a useful feature to me. Is it possible to set default device of a torch.rand tensor to based on a global device setting?

oliverholworthy:
>It's important to use data on the same device that the model will be on during serving when tracing the model

important or necessary? do you know why this has to be the case for the tracing to work correctly?

I can see [something in the TorchScript FAQ](https://pytorch.org/docs/stable/jit.html#frequently-asked-questions)

> recommended because the tracer may witness tensor creation on a specific device, so casting an already-loaded model may have unexpected effects. Casting the model before saving it ensures that the tracer has the correct device information.

Does that mean it has to be exactly the _same_ device, not only GPU vs CPU. (but the same GPU architecture and device count for example?)

karlhigley:
I suppose you could do `torch.set_default_tensor_type(torch.cuda.FloatTensor)`, but I think passing around a device argument is the recommended way to do it.

Issue #505, opened 2022-10-21
[Task] Standardize the model output format. 


sararb:
### Description

Currently, the T4Rec model outputs can be either: 

-  A tensor: representing the predictions returned by a single task. 
-  A dictionary of 3 tensors ({'loss': .., 'labels':..., 'predictions':....}}) representing the output of NextItemPredictionTask with `hf_format=True`
- A dictionary where the key is the task name and the value is the prediction tensor or the 3 tensors {'loss': .., 'labels':..., 'predictions':....}}

The objective of this task is to unify the output format returned by the Model and make sure the right information is returned based on whether we are in training, testing, or inference mode. 

**Additional context**
- This is needed for the multi-GPU support of Binaryclassification and Regression tasks. #487 

### Steps for refactoring: 

- [x] #542

<details><summary>Three tasks</summary>

-  Set the model to the inference mode by default  + ensure training script is not impacted + update failing unit-tests   **[Done]**

- Remove the hf-format from Next Item Prediction task + ensure training / inference are not impacted + update failing unit-tests   **[Done]**

- Replace `ignore_masking` with `testing` flag + ensure training / inference are not impacted + update failing unit-tests   **[Done]**
</details>

- [x] #544
i.e. compute loss inside the forward method + return dict with `{"loss", "labels" and "predictions"}` during training/evaluation mode. 
<details><summary>Four tasks</summary>

- Update base PredictionTask class:  **[Done]**
      ⁃	Add the `targets` argument to the forward method  
      ⁃	Move compute_loss() inside the forward method 
      ⁃	Return output (dict with the three keys or  torch.Tensor) based on training and testing flags
      ⁃	Update calculate metrics to pass the `targets` to the forward call 

- Update Head and Model classes to support the new convention in their forward method call  + calculate_metrics **[Done]**

- Update the fit method  **[in progress]**:  loss is computed inside the forward call + add flag `compute_metrics=True` to control whether to compute metrics during training or not.  Replace the `compute_loss`  call `loss = self.compute_loss(x, y)` by : 
```
outputs = self(x, y, training=True)
loss = outputs['loss']
if compute_metrics=True: 
    self.calculate_metrics(outputs['predictions'], outputs['labels'], mode='train', forward=False, call_body=False)
```

- Update the failing unit tests **[Open]** [ Blocked by Part 1]

</details>

- [x] #554

<details><summary>Three tasks</summary>

[...]

PR #507, opened 2022-10-21
Add save/load & input/output schema methods to T4Rec Model class 

### LINKED ISSUE #499: [Task] Transformers4Rec Serving using Merlin Schema
- [x] Saving Merlin input and output schema when saving the model

sararb:
Fixes #499

### Goals :soccer:
- Add save/load methods of T4Rec models using CloudPickle and following the protocol defined in [Merlin-Model](https://github.com/NVIDIA-Merlin/models/](https://github.com/NVIDIA-Merlin/models/pull/680/))

- Add an `input_schema` property to the T4Rec base class `Model`  that builds the model schema from the inputs modules of the heads and returns the merlin schema object. 

- Add an `output_schema` property to the T4Rec base class `Model` that builds the model schema from the predictions tasks specified in the heads and returns a merlin schema object with as many ColumnSchemas as the predictions tasks. 

- Add save/load methods to the T4Rec base class `Model` using `cloudpickle` and following the same protocol proposed in merlin models ([here](https://github.com/NVIDIA-Merlin/models/pull/680/))

- Add a `shape` property to input/output schema to provide the length/shape information of list features 

-  I used the code of this unit test in merlin [system](https://github.com/NVIDIA-Merlin/systems/blob/main/tests/integration/t4r/test_pytorch_backend.py#L60-L92) as a starting point to convert the input T4Rec schema to a merlin schema object. 

- The output schema is built based on the prediction tasks provided to the model. The stored information is: 
`name, int_domain, value_counts, is_list, shape, and is_ragged`. 

### Constraints

- The format of the T4Rec model outputs is not standardized and varies a lot based on the PredictionTask and some specific boolean flags such as `hf_format`. There is a working going on to simplify the output API format (#505) which will simplify the output of the model at inference (one prediction tensor is returned in the case of a single task learning or a dictionary of tensors where keys are the task name and values are the predictions tensors). 

- The output dictionary needs to be converted to a `NamedTuple` for PyTorch serving. 

### Testing Details :mag:
- Update model tests with testing input/output schemas. 
- Add a unit test `test_save_next_item_prediction_model`: saving/loading the model trained with the next item prediction task (the model used in the inference example)

### DIFF
--- a/tests/torch/model/test_model.py
+++ b/tests/torch/model/test_model.py
@@ -43,10 +43,10 @@ def test_simple_model(torch_tabular_features, torch_tabular_data):
     losses = model.fit(dataset, num_epochs=5)
     metrics = model.evaluate(dataset, mode="eval")
 
-    in_schema = model.input_schema(max_sequence_length=20)
+    in_schema = model.input_schema
     assert isinstance(in_schema, Core_Schema)
     assert set(in_schema.column_names) == set(inputs.schema.column_names)
-    out_schema = model.output_schema()
+    out_schema = model.output_schema
     assert isinstance(out_schema, Core_Schema)
     assert len(out_schema) == 1
     assert out_schema.column_names[0] == "target/binary_classification_task"
@@ -149,21 +149,14 @@ def test_model_with_multiple_heads_and_tasks(
     assert len(losses) == 5
     assert all(loss is not None for loss in losses)
 
-    # Get input schema
-    with pytest.raises(ValueError) as exc_info:
-        model.input_schema()
-        assert "You should specify the `max_sequence_length` of your list input" in str(
-            exc_info.value
-        )
-
-    in_schema = model.input_schema(max_sequence_length=20)
+    in_schema = model.input_schema
     assert isinstance(in_schema, Core_Schema)
     assert set(in_schema.column_names) == set(
         non_sequential_features_schema.column_names
         + torch_yoochoose_tabular_transformer_features.schema.column_names
     )
     # Get output schema
-    out_schema = model.output_schema()
+    out_schema = model.output_schema
     assert isinstance(out_schema, Core_Schema)
     assert set(out_schema.column_names) == set(output.keys())
 
@@ -234,11 +227,11 @@ def test_item_prediction_transformer_torch_model_from_config(
     assert out.size()[1] == task.target_dim
     assert len(out.size()) == 2
 
-    in_schema = model.input_schema(max_sequence_length=20)
+    in_schema = model.input_schema
     assert isinstance(in_schema, Core_Schema)
     assert set(in_schema.column_names).issubset(yoochoose_schema.column_names)
     # Get output schema
-    out_schema = model.output_schema()
+    out_schema = model.output_schema
     assert isinstance(out_schema, Core_Schema)
     assert len(out_schema) == 1
     assert out_schema.column_names[0] == "next-item"
@@ -331,12 +324,12 @@ def test_save_next_item_prediction_model(
     output = loaded_model(torch_yoochoose_like, training=False)
     assert isinstance(output, torch.Tensor)
 
-    in_schema = loaded_model.input_schema(max_sequence_length=20)
+    in_schema = loaded_model.input_schema
 
     assert isinstance(in_schema, Core_Schema)
     assert set(in_schema.column_names) == set(inputs.schema.column_names)
 
-    out_schema = loaded_model.output_schema()
+    out_schema = loaded_model.output_schema
     assert isinstance(out_schema, Core_Schema)
     assert len(out_schema) == 1
     assert out_schema.column_names[0] == "next-item"
--- a/tests/torch/model/test_model.py
+++ b/tests/torch/model/test_model.py
@@ -14,7 +14,12 @@
 # limitations under the License.
 #
 
+import os
+import tempfile
+
 import pytest
+import torch
+from merlin.schema import Schema as Core_Schema
 
 from transformers4rec.config import transformer as tconf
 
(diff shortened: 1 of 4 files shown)


transformers4rec/torch/model/base.py
```diff
@@ -514,6 +518,7 @@ def __init__(
         head_reduction: str = "mean",
         optimizer: Type[torch.optim.Optimizer] = torch.optim.Adam,
         name=None,
+        hf_format=True,
```
oliverholworthy, line 521: what does `hf_format` stand for? can we add a short note to the docstring. 

Can we move the docs for the init Parameters down from the class into the init method here.

sararb: The parameter is specific to NextITemPredictionTask and controls the output format required by the hugging face dependency. This confusing parameter will be removed in #505. In the meantime, I added a short description in the class docstrings.

transformers4rec/torch/model/base.py
```diff
@@ -678,6 +697,143 @@ def _get_name(self):
...
+                )
+            value_count = (
+                {"min": max_sequence_length, "max": max_sequence_length}
+                if is_list
+                else {"min": 1, "max": 1}
+            )
+            is_ragged = is_list and value_count.get("min", 0) != value_count.get("max", 0)
+            int_domain = {"min": column.int_domain.min, "max": column.int_domain.max}
+            shape = max_sequence_length if is_list else 1
+            properties = {"value_count": value_count, "int_domain": int_domain, "shape": shape}
```
oliverholworthy, line 730: In Merlin Systems, this `shape` is expected to include the batch dimension specified as -1 e.g. `[-1, 20]` for a list of size 20 or `[-1]` for a single value.

Perhaps we need to agree on the convention we're going for with this shape or maybe pull this shape into the Schema so that this can be checked and made consistent

tests/torch/model/test_model.py
```diff
@@ -200,6 +234,15 @@ def test_item_prediction_transformer_torch_model_from_config(
     assert out.size()[1] == task.target_dim
     assert len(out.size()) == 2
 
+    in_schema = model.input_schema(max_sequence_length=20)
+    assert isinstance(in_schema, Core_Schema)
+    assert set(in_schema.column_names).issubset(yoochoose_schema.column_names)
```
oliverholworthy, line 239: Why is this a subset and not the same?

(In this case `in_schema` is missing `{'ts/list', 'ts/first', 'ts/last', 'timestamp/age_days/list', 'day_idx', 'session_id', 'session_start', 'session_size'}`)

oliverholworthy: I tried this PR in the [Systems integration test with Transformers4Rec](https://github.com/NVIDIA-Merlin/systems/blob/v0.7.0/tests/integration/t4r/test_pytorch_backend.py) and got the following error trying to use the `model.input_schema` to define the input features, suggesting that maybe the missing cols are required?

```
RuntimeError: Number of dimensions of repeat dims can not be smaller than number of dimensions of tensor
```

transformers4rec/torch/model/base.py
sararb: Thank you @oliverholworthy for your review! Indeed adding the shape information to the schema properties was suggested by Karl when we were discussing what would be the best option to forward the information of non-ragged list features to merlin-systems. But sorry, I missed the standard format used in systems...
The shape information is specific to systems.  So I believe we can update the in/out schema in T4Rec to follow the same convention used in systems. 
Just to double-check, the shape of features is set in this [compute_dim](https://github.com/NVIDIA-Merlin/systems/blob/15074adf11bcc9bb9da375ec40ab4e431f1c93e9/merlin/systems/dag/ops/__init__.py#L18) function?

tests/torch/model/test_model.py
sararb: This is a subset of the original schema because the sequential input block in T4Rec (`TabularSequenceFeatures`) filters the inputs based on their tags: It keeps only variables tagged as `['list',  'categorical']` or `['list', 'continuous']`. So the features `'ts/list', 'ts/first', 'ts/last', 'day_idx', 'session_id', 'session_start', 'session_size'` are not lists, and  `'timestamp/age_days/list'` is tagged only as list.
In this unit test, the user can provide the whole schema (generated by NVTabular for example); and the T4Rec model is responsible for filtering the features. 

Can you share the code snippet to reproduce the bug and check what part of T4Rec is causing it?

oliverholworthy: It turns out that the error was caused by a different issue, not passing the correct shape input for some columns, `[-1, 1]` vs `[-1]` and not because of these missing features

transformers4rec/torch/model/base.py
```diff
@@ -543,7 +553,17 @@ def forward(self, inputs: TensorOrTabularData, training=True, **kwargs):
...
-            return outputs[list(outputs.keys())[0]]
+            outputs = outputs[list(outputs.keys())[0]]
+            if isinstance(outputs, dict):
+                # next-item prediction task with `hf_format = True``  returns a dictionary
+                # with three tensors: loss, predictions, labels. This needed for training
+                # and evaluation using the Trainer class.
+                # At inference, we just need the predictions tensors.
+                # TODO: We are simplifying the logic around `hf_format` in the multi-gpu
+                # support work.
+                if not training and not self.hf_format:
```
oliverholworthy, line 564: Can we remove `not training` from here so that the `hf_format` controls the output. That would make it work in systems without a model adapter wrapper class

transformers4rec/torch/model/base.py
```diff
@@ -678,6 +698,136 @@ def _get_name(self):
...
+
+            value_count = {"min": max_sequence_length, "max": max_sequence_length}
+            is_ragged = is_list and value_count.get("min", 0) != value_count.get("max", 0)
+            int_domain = {"min": column.int_domain.min, "max": column.int_domain.max}
+            batch_dim = [-1]
+            column_dim = [max_sequence_length] if is_list else [1]
+            properties = {
+                "value_count": value_count,
+                "int_domain": int_domain,
+                "shape": batch_dim + column_dim,
```
oliverholworthy, line 734: I think we can drop this shape from here since it's not used in systems for serving this

transformers4rec/torch/model/base.py
```diff
@@ -678,6 +698,136 @@ def _get_name(self):
...
+            else:
+                max_sequence_length = 1
+
+            value_count = {"min": max_sequence_length, "max": max_sequence_length}
+            is_ragged = is_list and value_count.get("min", 0) != value_count.get("max", 0)
+            int_domain = {"min": column.int_domain.min, "max": column.int_domain.max}
+            batch_dim = [-1]
+            column_dim = [max_sequence_length] if is_list else [1]
+            properties = {
+                "value_count": value_count,
```
oliverholworthy, line 732: we can drop this value count, since for serving got it working with variable sequence inputs.

oliverholworthy (approved):

PR #510, opened 2022-10-25
Cropped the table and addressed comments from previous PR

nzarif:
### Goals :soccer:

- Notify users about impact of multi_gpu training on metrics and how to address it (tune learning rate).
- Cropped the empty white space out of performance table.


docs/source/multi_gpu_train.md
```diff
@@ -51,4 +51,6 @@ We trained and evaluated a number of models using single GPU, `DataParallel` and
 
 <img src="_images/DP_DDP_perf.png" alt="Performance comparison of diffrerent training modes" style="width:800px;display:block;margin-left:auto;margin-right:auto;"/><br>
 
-These experiments used a machine with 2 <i>Tesla V100-SXM2-32GB-LS</i> GPUs.
\ No newline at end of file
+These experiments used a machine with 2 <i>Tesla V100-SXM2-32GB-LS</i> GPUs.
+
+<b>Note:</b> Using Data-Parallel training can impact metrics such at NDCG and recall. Therefore, Users are advised to tune the learning rate for better accuracy.
```
rnyak, line 56: Can we add some recommendations/more clarification about that? for example, if we increase the gpu number 2x, do we expect learning rate to be increased or reduced accordingly? 

can you make `Users` lower case? like `Therefore, users are ...`

nzarif: In most cases we should decrease learning rate as we increase number of GPUs but I did not observe any specific pattern in how learning rate must be changed. It differs based on model type, batch_size and etc. For example I saw a model where the same metrics could be achieved with 2 GPUs with the same learning rate. I also saw an example where for 2 GPUs, learning rate must be multiplied by 0.75.
I can add learning_rate to the table, but would that be enough? Should I mention any other parameter of the too? I have not modified all the different parameters when running the script, so I was not 100% sure if learning rate only depends on nr_gpu+model_type+batch_size or not.

nzarif: loss does not change much between single GPU, `DataParallel` and `DistributedDataParallel` but accuracy and NDCG changes considerably. It changes for both `DataParallel` and `DistributedDataParallel` assuming in each case, each GPU is receiving the same batch-size. This means each GPU is processing the same batch-size on each step so when more than 1 GPU is used, the effective batch-size processed at once is larger. But it does not translate to the same learning_rate producing same accuracy/NDCG/recall for both `DataParallel` and `DistributedDataParallel` and I believe the reason lies in the different approach these modes have for aggregation after processing a batch.

nzarif: I did play around with the learning_rate for a couple of models and I could find learning_rates that gave same metrics for each of the modes. I did not record the learning_rates back then but if you would like to, I can retry doing that and come up with some examples of learning_rates that produce similar metrics for each of the training modes (single GPU, `DataParallel` and `DistributedDataParallel`).

karlhigley:
@rnyak @gabrielspmoreira Is this PR still relevant or has it been superseded by other work? Should we push it forward toward merging or close it as outdated?

rnyak (approved):

Issue #509, opened 2022-10-25
[BUG] MeanReciprocalRankAt created on CPU


tetelias:
### Bug description

When calculating MeanRecipricolRankAt metric in https://github.com/NVIDIA-Merlin/Transformers4Rec/blob/main/transformers4rec/torch/ranking_metric.py:313 `(torch.arange(k) + 1)` is created on CPU, so when labels are on GPU, it throws an error.

I've replaced line 313 with 
```
            values, _ = (topk_labels[:, :k] / (torch.arange(k) + 1).to(
                device=scores.device)).max(dim=1)
```

### Steps/Code to reproduce bug

It's straightforward so no clarification needed.

### Expected behavior

Created tensor should be on the same device as labels.

### Environment details
- Transformers4Rec version: 0.1.13
- Platform: Ubuntu 18.04
- Python version: 3.8.8
- Huggingface Transformers version: 4.23.0
- PyTorch version (GPU?): 1.11 GPU
- Tensorflow version (GPU?):

PR #513, opened 2022-11-01
fixing name and bug in MeanReciprocalRankAt

tetelias:
Fixes # (issue)
[BUG] #509
### Goals :soccer:

Fixes the problem with one of the tensors always created on CPU even if data is processed with GPU.
Also name is corrected.

Tensor is placed on the same device as previously created one.
### Testing Details :mag:

Other metrics are not tested so I don't have a guidance for the process and the bug/naming is rather straightforward so there is no tests.

### DIFF
--- a/tests/torch/test_ranking_metrics.py
+++ b/tests/torch/test_ranking_metrics.py
@@ -17,7 +17,7 @@
 import pytest
 import torch
 
-from transformers4rec.torch.ranking_metric import MeanRecipricolRankAt
+from transformers4rec.torch.ranking_metric import MeanReciprocalRankAt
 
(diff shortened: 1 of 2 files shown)


rnyak:
@tetelias thanks for the PR. can you please reformat the files. you can install isort, black and flake8 and run them to fix any pre-commit issues before committing.  or you can do the following before commiting:
```
pip install pre-commit
pre-commit install
```

sararb (approved):

rnyak:
closing since we fix that here https://github.com/NVIDIA-Merlin/Transformers4Rec/pull/522

PR #533, opened 2022-11-15
Add docs requirements to extras list in setup.py

oliverholworthy:
Follow-up to #527

Docs requirements were moved out into a new file.

For the docs re-build which installs the package as `.[all]`, adding this as a `docs` requirements extra.


mikemckiernan (approved):
lgtm, thanks very, very much.

Issue #487, opened 2022-09-08
[Task] Support multi-GPU training for all prediction tasks


sararb:
### Description
Support multi-GPU training for any prediction tasks provided by T4Rec. This can be done in two different ways:

### Goals
- **Using the custom fit method:**
- [ ] Support wrapping the model in DistributedDataParallel or DataParallel within the fit method
- [ ] Check if calling `self.wrapped_model.module.compute_loss` maintains the distributed mode (All GPUs are used)

- **Using the custom the Trainer class**
- [ ] Update all prediction tasks to return a dictionary containing the loss score as expected by the HF trainer. Similar to what we have in NextItemPredictionTask ([here](https://github.com/NVIDIA-Merlin/Transformers4Rec/blob/21a2a836a30610746bc8788ff926bad74fd00289/transformers4rec/torch/model/prediction_task.py#L247))
- [ ] Ensure the custom [evaluation](https://github.com/NVIDIA-Merlin/Transformers4Rec/blob/21a2a836a30610746bc8788ff926bad74fd00289/transformers4rec/torch/trainer.py#L346) loop is working for all T4R predictions task. Especially, enabling [the top-k scores calculcation](https://github.com/NVIDIA-Merlin/Transformers4Rec/blob/21a2a836a30610746bc8788ff926bad74fd00289/transformers4rec/torch/trainer.py#L477) only for NextItemPrediction of MultiClassification use-cases.

PR #540, opened 2022-11-19
Add music-streaming synthetic data to test the support of all predictions tasks with the Trainer class

sararb:
- This PR adds a synthetic music-streaming dataset generated from the merlin models data package and a unit test of the Trainer using this dataset. 
- The motivation is to add a testing dataset that has both regression and binary classification targets. This is needed to complete the tests of the task `multi-GPU support for all prediction tasks`(#487 [#708](https://github.com/NVIDIA-Merlin/Merlin/issues/708))


nzarif (approved):
Thanks. For the TODO part I will add Task types and metrics as test parameters.

PR #543, opened 2022-11-19
Refactoring part1- flags modification

### LINKED ISSUE #542: Part 1: Simplify the existing API
This part includes 3 tasks:

- Set the model to the inference mode by default + ensure training script is not impacted + update failing unit-tests

- Remove the hf-format from Next Item Prediction task + ensure training / inference are not impacted + update failing unit-tests

- Replace ignore_masking with testing flag + ensure training / inference are not impacted + update failing unit-tests

P.S: This PR removes the `hf_format` only for `NextItemPredictionTask`. Other tasks such as `BinaryClassificationTask` and `RegressionTask` still use `hf_format` as removing the flag for them at this stage does not make sense for them, they use the `PredictionTask.forward()` only for inference and use `compute_loss()` for training and eval.

nzarif:
Fixes #542 
### Goals :soccer:
- Set the model to the inference mode by default + ensure training script is not impacted + update failing unit-tests
- Remove the hf-format from Next Item Prediction task + ensure training / inference are not impacted + update failing unit-tests
- Replace ignore_masking with testing flag + ensure training / inference are not impacted + update failing unit-tests

These changes help with standardizing T4Rec API and enable using HuggingFace `Trainer` for training all `PredictionTask`s within T4Rec.

### Testing Details :mag:
Run the torch unit tests and you will see they pass. You can also try with [integration tests](https://github.com/NVIDIA-Merlin/Transformers4Rec/blob/main/ci/test_integration.sh).


transformers4rec/torch/model/base.py
```diff
@@ -545,12 +554,19 @@ def __init__(
         self.optimizer = optimizer
         self.hf_format = hf_format
```
sararb, line 555: Can we remove setting the `hf_format` from the model's init method? (from the model's forward method and the docstring as well) This was used to set the `hf_format` of `NextItemPredictionTask` to False after training to prepare the model for inference. But with the new refactoring introduced in this PR, the model will always return a tensor at inference (which is also the default mode when calling `model(inputs)`)

sararb (commented):
The PR looks good to me! I also pulled it into my local machine and ran the ci tests all the scores are matching the ones we get from the main branch. 

I have just left one comment to remove setting the `hf_format` in the base `Model` class.

sararb (approved):

Issue #524, opened 2022-11-09
[Task] Use the same logic of the new merlin DataLoader to set the features types in T4Rec


sararb:
### Description

- The previous nvtabular data loader was always returning variables with the dtype `float32` and `int32`. We followed the same standard format in T4Rec which requires inputs to be of `float32/int32`. 

- The new data loader changed the convention of always returning `float32/int32` to return the same type used in the source data being loaded (e.g. from a parquet file).  

- This change in data loader causes an error in T4rec when the parquet file was containing features with `float64` dtypes ( more details [here](https://github.com/NVIDIA-Merlin/NVTabular/issues/1698)).

- This feature request aims to extend T4Rec to use the same types used in the source data being loaded

PR #547, opened 2022-11-22
Uses merlin-dataloader package

edknv:
Fixes https://github.com/NVIDIA-Merlin/dataloader/issues/16

### Goals :soccer:
- Replace nvtabular dataloader with the new merlin-dataloader.

- NVTabular was updated a while ago to use the new merlin-dataloader package, and and as a result, T4rec has already been using merlin-dataloader via NVTabular. The new dataloader had a number of changes on dtypes, etc. and most of the work necessary to make it work with T4rec had already been completed by @sararb in #536, #539, #541, and #523. This PR proposes to create a loader class that inherits from `merlin.loader.torch.Loader`. Since NVTabular is using `merlin.loader.torch.Loader` under the hood, replacing `NVTabularDataLoader` with `MerlinDataLoader` pretty much amounts to a name change. There should be no difference in functionality, but this change has the advantage that we make the dependency explicit and it becomes less susceptible to breaking changes in NVTabular.
- See https://github.com/NVIDIA-Merlin/dataloader/issues/16 for more details on dtypes and the new dataloader using all the columns in the dataset schema. See also #524 for potential future work. It is not clear if we should commit to more work on T4Rec when session-based models have been implemented in Merlin Models and there is plan to support Torch in Models.

### Testing Details :mag:
`tox -e test-gpu`
`./ci/test_integration.sh`


edknv:
rerun tests

edknv:
The naming convention in dataloader has been changed from `merlin.loader` to `merlin.dataloader`: https://github.com/NVIDIA-Merlin/dataloader/pull/57. I'll update the PR.

merlin_standard_lib/utils/misc_utils.py
```diff
@@ -236,4 +236,42 @@ def validate_dataset(paths_or_dataset, batch_size, buffer_size, engine, reader_k
             reader_kwargs["batch_size"] = buffer_size
     else:
         reader_kwargs["part_mem_fraction"] = buffer_size
+
     return Dataset(files, engine=engine, **reader_kwargs)
+
+
+def _augment_schema(
```
sararb, line 243: In the T4Rec data loader, we have a filtering step happening [here](https://github.com/NVIDIA-Merlin/Transformers4Rec/blob/776af7bad969fb408e623d1a3b832e8d5ee337a4/transformers4rec/torch/utils/data_utils.py#L228) to filter out the columns that are not used by the model before passing the dataset to the loader. This step is required to make sure that the data loader is returning only the input features required by the model  . 

so perhaps we can add this filtering step to the `_augment_schema` method to keep all the schema processing logic in one place.

merlin_standard_lib/utils/misc_utils.py
```diff
@@ -236,4 +236,42 @@ def validate_dataset(paths_or_dataset, batch_size, buffer_size, engine, reader_k
...
+    for label in cats or []:
+        schema[label] = schema[label].with_tags(Tags.CATEGORICAL)
+    for label in conts or []:
+        schema[label] = schema[label].with_tags(Tags.CONTINUOUS)
+
+    # Set the appropriate properties for the sparse_names/sparse_max/sparse_as_dense
+    for col in sparse_names or []:
+        cs = schema[col]
+        properties = cs.properties
+        if sparse_max and col in sparse_max:
```
sararb, line 266: From what I understand: `sparse_max` is a dict where keys are the sparse feature names and values are the max-length to use for padding, and `sparse_names` is the list of sparse feature names. So is there a use case where the list `sparse_names` is different from the list of the ` sparse_max` keys?

sararb (commented):
Thank you for your PR! This makes the t4rec loader much easier the maintain with the merlin loader package. 
I have just left two comments about the `_augment_schema` method

merlin_standard_lib/utils/misc_utils.py
edknv: This logic was ported over from nvtabular, but I think there are some cases where `sparse_names` and `sparse_max` might be different (or `sparse_max` is under-specified). For example, when we want the value count of a feature to be [unbounded](https://github.com/NVIDIA-Merlin/core/blame/78f1f0b0952fd14b76913b0dd258565c06694abe/merlin/schema/schema.py#L42), we would leave out `sparse_max` but keep `sparse_names`.

sararb (approved):
Thank you @edknv for the updates! I pulled your PR locally and run the ci tests with the latest core and dataloader packages. Everything runs fine without any issues.

Issue #506, opened 2022-10-21
[QST] Unable to replicate evaluation metrics when using ignore_masking.


TassaraR:
# ❓ Questions & Help

## Details
`transformers4rec==0.1.13`

I've been trying out t4r for a while and I decided to try and replicate the evaluation metrics on my own by performing offline predictions.

I've been using a model architecture similar to one in the provided examples and similar features.

Model:
```python3
inputs = tr.TabularSequenceFeatures.from_schema(
        schema,
        max_sequence_length=36,
        aggregation='concat',
        continuous_projection=64,
        d_output=64,
        masking="mlm",
)

# Define XLNetConfig class and set default parameters for HF XLNet config  
transformer_config = tr.XLNetConfig.build(
    d_model=64, n_head=4, n_layer=2, total_seq_length=36
)
# Define the model block including: inputs, masking, projection and transformer block.
body = tr.SequentialBlock(
    inputs, tr.MLPBlock([64]), tr.TransformerBlock(transformer_config, masking=inputs.masking)
)

# Defines the evaluation top-N metrics and the cut-offs
metrics = [NDCGAt(top_ks=[5, 10, 20], labels_onehot=True),  
           RecallAt(top_ks=[5, 10, 20], labels_onehot=True),
           AvgPrecisionAt(top_ks=[5, 10, 20], labels_onehot=True),
           PrecisionAt(top_ks=[5, 10, 20], labels_onehot=True)
          ]

# Define a head related to next item prediction task 
head = tr.Head(
    body,
    tr.NextItemPredictionTask(weight_tying=True, hf_format=True, 
                              metrics=metrics),
    inputs=inputs,
)

# Get the end-to-end Model class 
model = tr.Model(head)

train_args = T4RecTrainingArguments(data_loader_engine= 'nvtabular', 
                                    dataloader_drop_last = False,
                                    gradient_accumulation_steps = 1,
                                    per_device_train_batch_size = 256, 
                                    per_device_eval_batch_size = 32,
                                    output_dir = "./tmp", 
                                    learning_rate=0.0005,
                                    lr_scheduler_type='cosine',
                                    num_train_epochs=5,
                                    max_sequence_length=36, 
                                    report_to = [],
                                    logging_steps=200,
                                    no_cuda=False)

trainer = Trainer(
    model=model,
    args=train_args,
    schema=schema,
    compute_metrics=True,
)
```

When I evaluate the metrics by using:
```python3
[...]

PR #551, opened 2022-11-23
Fix discrepancy between evaluation and inference modes 

### LINKED ISSUE #525: [BUG] Model's performance discrepancy between evaluation and inference modes
### Bug description

This BUG is a summary of the thread discussion happening in #506.  The main issue is the discrepancy of scores when applying the model using the masking to replace the last interaction with [MASK] embeddings and applying the same model without using the masking replacement. Instead of masking, the last interaction of the input sequence is explicitly set to `0` (to mimic the real-life inference scenario)

Given the input sequence, a subset of positions is replaced by special embeddings [MASK], and the model is trained to recover the original item-id from the corrupted input sequence.   During inference, we don't rely on data corruption (using [MASK]). Instead, we apply the model to the original test input sequence. Hence, we observe a discrepancy in scores because the model learned latent information from [MASKING] embeddings which is not used in the original test data. 

1. I summarized the code shared in thread #506 in this [gist](https://gist.github.com/sararb/7e61b286aa612a2d707a59832bd948ff#file-replicate_t4rec_eval_metrics-py).
2. The code can be added to the notebook `02-End-to-end-session-based-with-Yoochoose-PyT` after the cell 10 to get the different test scores.

### Expected behavior
- The performance scores should match between the training/evaluation mode where data is corrupted with the [MASKING] embeddings and the test/inference mode where the masking is not applied to the original data. 

### Additional context
Some possible solutions to fix the discrepancy are: 
- Extend the test data with a dummy last interaction and use the model with `ignore_masking=False` to replace the last position with the [MASK] embeddings. 
- Replace the trainable masked embeddings tensor by the embeddings of the padding-idx `0`

sararb:
Fixes #525 #506 

### Goals :soccer:
- [x] Run experiments to understand what is causing the discrepancy issue
- [x] Implement a fix of `MaskedLanguageModeling` to make sure inference and evaluation scores are aligned 
- [x] Updating the CI script to compute the simulated inference recall@10 score.
- [x] Fix the `prediction_step` of the Trainer class when `testing=False` (i.e. inference mode) 

### Experiment analysis: 
I ran different experiments to understand the cause of the issue. The main takeaway from the experiments is that the transform-based model trained with MLM is sensitive to the target's positional encoding. For more details about the results of the experiments, you can click on the following sections: 

<details><summary> Problem definition </summary> 
The current implemented strategy in Transformers4Rec to train, evaluate, and deploy transformer-based models trained with Masked language modeling is as follows: 

image

* From the figure above, we could observe that the main difference between Train/Eval and Inference modes are: 
  * Evaluation uses the hidden representation of [MASK] as input to the softmax layer while during inference we use the hidden representation of the last interaction instead. 
  * The hidden representation selected during `evaluation` includes the actual position of the target we would like to predict while during inference the hidden representation is related to the previous position (i.e target position - 1)

From these two observations, we drew the following questions: 

  - **Question 1:** What if we used the representation of the first PAD (‘0’) observed after the last interaction during inference to take into account the actual target position? 
  - **Question 2:** What if we used the same embeddings for MASK and PAD  (‘0’) positions? 
  → The rationale is to align the target hidden representation vector used during training and inference modes. 
  - **Question 3:** Instead of using the first padded `0` after the last interaction, what if we extend the test sequence with an additional position [MASK] after the last interaction?
  → As using the first `0` position in the current sequence will not work for full sequences with no padded positions.
</details>

<details><summary> Experiment setup </summary> 

To answer questions defined in the `Problem definition` section, we ran the following experiment:

  - Scripts: The integration test based on the t4rec paper reproducibility [script](https://github.com/NVIDIA-Merlin/Transformers4Rec/blob/main/examples/t4rec_paper_experiments/t4r_paper_repro/transf_exp_main.py) and the end-to-end example [notebook](https://github.com/NVIDIA-Merlin/Transformers4Rec/blob/main/examples/end-to-end-session-based/02-End-to-end-session-based-with-Yoochoose-PyT.ipynb).

  - Datasets: ecomrees46 with only item-id, Yoochoose with item-id, and Yoochoose with additional item features. 

  <details><summary>-Evaluation strategies: </summary>  

1. `Trainer.evaluate()`: Using the offline strategy, we mask the input sequence's last interaction and apply the trained model to predict the masked item-id. 
  Original: [[2, 34, 5, 50, 2, 0],  [2, 34, 5, 50, 2, 4] ]: 
  Inputs:[ [2, 34, 5, 50, 2, 0],  [2, 34, 5, 50, 2, 4] ]
  Masking:[ [2, 34, 5, 50, [MASK], 0],  [2, 34, 5, 50, 2, [MASK] ]
  Label: [2, 4]
2. `Simulated inference`: Extract the last interaction as labels, replace it with `0` in the original input sequence then apply the model without masking. 
  Original: [[2, 34, 5, 50, 2, 0],  [2, 34, 5, 50, 2, 4] ]: 
  Inputs:[ [2, 34, 5, 50, 0, 0],  [2, 34, 5, 50, 2, 0] ]
 	  Masking:[ [2, 34, 5, 50, 0, 0],  [2, 34, 5, 50, 2, 0] ]
  Label: [2, 4]
3.` Simulated inference with a `dummy` interaction`: Replace the label position with a dummy variable `100` then apply the model with masking. 
  Original: [[2, 34, 5, 50, 2, 0],  [2, 34, 5, 50, 2, 4] ]: 
  Inputs:[ [2, 34, 5, 50, 100, 0],  [2, 34, 5, 50, 2, 100] ]
 	  Masking:[ [2, 34, 5, 50, [MASK], 0],  [2, 34, 5, 50, 2, [MASK]]
  Label: [2, 4]
</details>
 </details>
 <details><summary>The experiment results:  </summary>  
  
image
</details>
  <details><summary>Results analysis:</summary>
 
* We observe that the baseline (current implementation) is strongly underperforming during the inference mode which implies that the current strategy used when `ignore_masking=True` (`testing=False` in the latest T4Rec API) is not working. 

* In question 1, the scores between the evaluation and the simulated inference are more aligned, suggesting that taking into account the target's true position is critical for performance. However, this solution is limited to test sequences with a number of interactions strictly lower than the maximum length. 

* In question 2, we wanted to test whether we should use the PAD value `0` to represent the masked targets instead of a [MASK] trainable embeddings. We couldn’t implement this strategy in the case where side information is available (extra item/session features are fed to the model in addition to the item ids) as it would require computing the embeddings of the PAD position for all features which is compute-intensive and very tricky to set for continuous variables. We observe that the obtained results are worse than the baseline, which suggests that the model is learning to differentiate between the padded positions and the masked ones.

* Combining the solutions of question 1 and question 2  confirmed the observation of question 2 as the gap between the inference and evaluation mode is larger than the one observed in question 1. 

* From these results, we implemented the solution of question 3 that extends the input sequence, during inference, with a [MASK] position right after the last seen interaction and uses the related hidden representation given by the transformer layer as input to the softmax layer to get the final logits scores. 

</details>

- Update the methods  `_compute_masked_targets` and `apply_mask_to_inputs`  of the masking class MLM to take into account the actual target position. This is done by extending the input sequence with one additional position using the [MASK] embeddings vector.

- Making sure that inputs are not modified during inference for other masking strategies (CLM and PLM) 

- Update the `prediction_step` method of the Trainer to align with the new refractory introduced in #542  

- I also had to set the "total_seq_length" argument of the transformer configuration to  ```total_seq_length=total_seq_length+ 2``` in order to take into account the additional position used at inference. 
### Testing Details :mag:
- Remove `--max_steps 20` in the ci command line that runs the model `XLNet - MLM - item_id`
- Add an assertion check to the CI script to make sure evaluation and simulated inference scores are matching. 

<details><summary>Click here to preview the CI results for the `main` and `fix-inference` branches</summary>

image

image
</details>

### Not included in this PR:

- Analysis of PLM task at inference and its related fix


examples/t4rec_paper_experiments/t4r_paper_repro/transf_exp_main.py
```diff
@@ -182,6 +185,69 @@ def main():
...
+        assert np.isclose(recall_10, results_over_time[2]["eval_/next-item/recall_at_10"], rtol=0.1)
+
+
+def recall(predicted_items: np.ndarray, real_items: np.ndarray) -> float:
+    idx = 0
+    recalls = np.zeros(len(predicted_items), dtype=np.float64)
+    for real, pred in zip(real_items, predicted_items):
+
+        real = real[real > 0]
+        pred = pred[pred > 0]
```
gabrielspmoreira, line 236: I understand `pred` is a 2D tensor with the top-k predicted ids, and eventually we might have a predicted item 0. Not sure has this op `pred[pred > 0]` will work in such case. I think it is better to align the filter with the `real`, (e.g. `pred[real > 0]`

examples/t4rec_paper_experiments/t4r_paper_repro/transf_exp_main.py
```diff
@@ -182,6 +185,69 @@ def main():
...
+    # Verify that the recall@10 from train.evaluate() matches the recall@10 calculated manually
+    if not isinstance(input_module.masking, t4r.masking.PermutationLanguageModeling):
+        # TODO fix inference discrepancy for permutation language modeling
+        assert np.isclose(recall_10, results_over_time[2]["eval_/next-item/recall_at_10"], rtol=0.1)
+
+
+def recall(predicted_items: np.ndarray, real_items: np.ndarray) -> float:
+    idx = 0
+    recalls = np.zeros(len(predicted_items), dtype=np.float64)
+    for real, pred in zip(real_items, predicted_items):
```
gabrielspmoreira, line 233: This could be done in a vectorized way, as we do in the `RecallAt` metric. Something like

```python
predicted_correct_positions = np.equal(predicted_items, real_items).astype(np.int32)
predicted_correct_sum = np.sum(predicted_correct_positions, axis=1)
num_relevant = real_items.shape[1]
valid_rows = (num_relevant != 0)
predicted_correct_sum = predicted_correct_sum[valid_rows]
num_relevant = predicted_correct_sum[valid_rows]
recall_per_row = predicted_correct_sum / num_relevant
return np.mean(recall_per_row)
```

transformers4rec/torch/masking.py
```diff
@@ -338,21 +373,42 @@ def _compute_masked_targets(self, item_ids: torch.Tensor, training=False) -> Mas
...
+            # At inference we extend the input with a [MASK] element at the first padded position
+            # to take into account the positional encoding of the target
+            labels = torch.full(
+                (item_ids.shape[0], item_ids.shape[1] + 1),
+                self.padding_idx,
+                dtype=item_ids.dtype,
+                device=item_ids.device,
+            )
+            last_item_sessions = non_padded_mask.sum(dim=1)
+            labels[rows_ids, last_item_sessions] = item_ids[rows_ids, last_item_sessions - 1]
```
gabrielspmoreira, line 405: Is this inference mode? From `last_item_sessions - 1` it seems we are masking the last item of the sequence, which is expected for evaluation mode, but not for inference mode, right?

transformers4rec/torch/masking.py
```diff
@@ -402,6 +458,34 @@ def _compute_masked_targets(self, item_ids: torch.Tensor, training=False) -> Mas
 
         return MaskingInfo(mask_labels, labels)
 
+    @docstring_parameter(flags_parameters_docstrings=TRAINING_TESTING_FLAGS_DOCSTRING)
+    def apply_mask_to_inputs(
+        self, inputs: torch.Tensor, schema: torch.Tensor, training=False, testing=False
```
gabrielspmoreira, line 463: Rename `schema` to `mask_schema` too as you did above, as `schema` is a "reserved" word in Merlin and here the schema is a tensor with the mask, which might confuse the user

examples/t4rec_paper_experiments/t4r_paper_repro/transf_exp_main.py
sararb: I see your point! so the main difference with `RecallAt` is that the labels are not one hot encoded (real_items is a 1-d vector of ids) and the `predicted_items` are already sorted (2-d vector of shape (bs, top_k)). But, following your suggestion, I updated the recall function to do the check `label[i] is in predicted_items[i] for i in range(n_rows)` in a vectorized way.

transformers4rec/torch/masking.py
sararb: I renamed it to `mask_schema` in this specific method. But there are a set of methods/classes in `masking`  (not part of this PR scope) that uses the name `schema` instead of `mask_schema` so perhaps we can have a followup task that standardizes the masking code to use `mask_schema` instead of `schema` to refer to the boolean mask vector.

transformers4rec/torch/masking.py
sararb: Yes, it is the inference mode. Here the `labels` are of shape `[bs, max_length+1]` while the input `item_ids` is of shape `[bs, max_length]`. 
The logic here is to set the label at position `last_item_sessions` (which is the position immediately after the last interacted item) with the last interacted item id given by `item_ids[rows_ids, last_item_sessions - 1]`. So that we can define the boolean mask using the check  ```mask_labels = labels != self.padding_idx``` .  So, the resulting mask will have `True` values only for the position `mask[rows_ids, last_item_sessions]`

gabrielspmoreira: I see @sararb . Makes sense. Thanks

examples/t4rec_paper_experiments/t4r_paper_repro/transf_exp_main.py
```diff
@@ -182,6 +185,63 @@ def main():
...
+def recall(predicted_items: np.ndarray, real_items: np.ndarray) -> float:
+    bs, top_k = predicted_items.shape
+    valid_rows = real_items != 0
+
+    # reshape predictions and labels to compare
+    # the top-10 predicted item-ids with the label id.
+    real_items = real_items.reshape(bs, 1, -1)
+    predicted_items = predicted_items.reshape(bs, 1, top_k)
+
+    num_relevant = real_items.shape[-1]
```
gabrielspmoreira, line 239: I think you need to filter `num_relevant[valid_rows]` as you do with `predicted_correct_sum[valid_rows]`

gabrielspmoreira (approved):

PR #553, opened 2022-11-24
Adding Reference to Merlin Models for TensorFlow

bschifferer:


README.md
```diff
@@ -17,6 +17,8 @@ Transformers4Rec supports multiple input features and provides configurable buil
 
 You can build a fully GPU-accelerated pipeline for sequential and session-based recommendation with Transformers4Rec and its smooth integration with other components of [NVIDIA Merlin](https://developer.nvidia.com/nvidia-merlin):  [NVTabular](https://github.com/NVIDIA-Merlin/NVTabular) for preprocessing and [Triton Inference Server](https://github.com/triton-inference-server/server).
 
+TensorFlow Support: Transformer-based architectures in TensorFlow are added to the [Merlin Models](https://github.com/NVIDIA-Merlin/models) library.
```
mikemckiernan, line 20: My $0.02 is to remove the "TensorFlow Support:" part.

Alternatively, maybe a heading with "Related Repositories" and add the sentence under that heading.  I make that sugg only because I sense that you want to redirect our customers because they might expect to find TensorFlow support in this repo.

mikemckiernan (approved):

README.md
bschifferer: @mikemckiernan Yes - my suggestion is that we cross-link the repository. If a TensorFlow users gets to Transformer4Rec and sees only PyTorch - what will he/she do? Should he/she not be referenced to Merlin Models?

gabrielspmoreira, line 20:
```suggestion
TensorFlow Support: HuggingFace Transformer-based architectures in TensorFlow are added to the [Merlin Models](https://github.com/NVIDIA-Merlin/models) library.
```

PR #558, opened 2022-12-02
Add an example of a pre-trained embeddings support to `TabularSequenceFeatures` and  `TabularFeatures`

sararb:
The input block `TabularFeatures` have a parameter `text_embedding_module` to allow the user to define a custom TextEmbedding module that returns pre-trained vectors for given text-encoded inputs.  

This quick PR exposes the argument in `TabularSequenceFeatures` and also defines a unit test with a dummy ` NormalizeTextembeddings` module to showcase the functionality.  

related RMPs:

- https://github.com/NVIDIA-Merlin/Merlin/issues/211
- https://github.com/NVIDIA-Merlin/models/issues/1043


tests/torch/features/test_sequential.py
```diff
@@ -208,3 +208,46 @@ def test_sequential_and_non_sequential_tabular_features(yoochoose_schema, torch_
...
+
+    import torch
+
+    from transformers4rec.torch.features.base import InputBlock
+
+    class NormalizeTextembeddings(InputBlock):
+        def __init__(self, features: List[str], **kwargs):
+            """
+            Parameters:
+                features: Names of text features
```
rnyak, line 224: does that mean the column name of the text input feature?

tests/torch/features/test_sequential.py
```diff
@@ -208,3 +208,46 @@ def test_sequential_and_non_sequential_tabular_features(yoochoose_schema, torch_
...
+
+
+def test_sequential_tabular_features_with_text_module(yoochoose_schema, torch_yoochoose_like):
+    from typing import List
+
+    import torch
+
+    from transformers4rec.torch.features.base import InputBlock
+
+    class NormalizeTextembeddings(InputBlock):
```
rnyak, line 220: can we add one liner docstring to define what this class do?  to explain why text embeddings are required to be normalized?  we expect the user to define such class when they want to use the text embeddings right?

transformers4rec/torch/features/tabular.py
```diff
@@ -175,7 +175,7 @@ def from_schema(  # type: ignore
         output = cls(
             continuous_module=maybe_continuous_module,
             categorical_module=maybe_categorical_module,
-            text_embedding_module=None,
+            text_embedding_module=kwargs.pop("text_embedding_module", None),
```
rnyak, line 178: how is `text_embedding_module` propogated later on in embedding layer? do we concat them? how do we create the lookups?

tests/torch/features/test_sequential.py
sararb: yes, I updated the description.

tests/torch/features/test_sequential.py
sararb: I added a docstring explaining what the class is doing. The objective of this unit test is to showcase how to add a custom text-embeddings module to the T4Rec input block. So it assumes the user will define his custom `Textembeddings(InputBlock)` based on his dataset and his use-case.

transformers4rec/torch/features/tabular.py
sararb: It will be an additional module to the `Embeddings` and `Continuous` layer. So the module will expect as inputs the dictionary of raw features and should include all the processing logic needed to represent text features in its forward method. (such as filtering text features, performing embedding lookup (or not), and normalization). The output of this module will be aggregated with the output of `Embeddings` and `Continuous` layers based on the `aggregation` specified in `tr.TabularSequenceFeatures.from_schema(...)`

gabrielspmoreira: As the `text_embedding_module` is not specific to text, what about renaming it to something like `custom_module`?

tests/torch/features/test_sequential.py
rnyak: does that have to be text features? or we can give a more generalized description like `name of the external features we have pre-trained embeddings for`

PR #564, opened 2022-12-06
[WIP] Extend trainer class to support all T4Rec prediction tasks 

### LINKED ISSUE #554: Part 3: ADD support of training regression and binary classification to the Trainer class
Tasks included in this part:

- Analyze and verify the logical correctness of calling `calculate_metrics` from `Model.fit()` at this [line](https://github.com/NVIDIA-Merlin/Transformers4Rec/blob/4b1a6f1dfa99d4c5c6fe2b525d5ddae3d52103ff/transformers4rec/torch/model/base.py#L676).

- Add a simple script for evaluation of metrics and results when using `Trainer` for `BinaryClassification` and `Regression` such that they can be used for CI purposes like [this one](https://github.com/NVIDIA-Merlin/Transformers4Rec/blob/4b1a6f1dfa99d4c5c6fe2b525d5ddae3d52103ff/ci/test_integration.sh) that is being used for `NextItemPrediction`.
- Version control fixes:
_ Rebase with refactor_part2 and main to get synthetic music data testing for BC and regression.
_ Pull the changes of fix-inference branch to get latest `predict_step` and the updated ci tes
- Update the testing/training flags inside the `predict_step` and `eval_loop` and test they are working correctly.
- Move the top-k logic (specific to the next item prediction task) to a function of the Trainer class and call it inside the eval_loop only when the item-prediction task is used.
- Customize train_step to get the targets for the predictions tasks:
_ iterate over the dataloader in [training_loop](https://github.com/huggingface/transformers/blob/afce73bd9d891b55dcb8d4d875d17718ffa01ff0/src/transformers/trainer.py#L1769) using input that contains labels/targets added to inputs.
_ Call [training_step](https://github.com/huggingface/transformers/blob/afce73bd9d891b55dcb8d4d875d17718ffa01ff0/src/transformers/trainer.py#L2528) inside training loop and make sure inputs encompassed with labels are passed to it.
_ Call [compute_loss()](https://github.com/huggingface/transformers/blob/afce73bd9d891b55dcb8d4d875d17718ffa01ff0/src/transformers/trainer.py#L2576) from [training_step()](https://github.com/huggingface/transformers/blob/afce73bd9d891b55dcb8d4d875d17718ffa01ff0/src/transformers/trainer.py#L2528) and make sure labels arrive there with inputs.
_ Make sure the [model()](https://github.com/huggingface/transformers/blob/afce73bd9d891b55dcb8d4d875d17718ffa01ff0/src/transformers/trainer.py#L2586) is called with correct arguments passed to it.
- Use HF [nested_detach()](https://github.com/huggingface/transformers/blob/335f9bcd3457d2c4c9c6d9aeb2dae06ab20a7b56/src/transformers/trainer_pt_utils.py#L162) inside [T4Rec.Trainer.prediction_step()](https://github.com/NVIDIA-Merlin/Transformers4Rec/blob/ec92eb6500dad7a8e6222523ff54f89c80e196e6/transformers4rec/torch/trainer.py#L340) to handle cases when we have dict instead of Tensor. This will make sure corresponding inputs and targets are placed on the same device when using multiple GPUs.
- Add a unit test for testing the trained with binary and/or regression tasks.
- Explain the support of BC and regression tasks in the documentation of the Trainer class.

sararb:
Fixes #554 #505 

### Goals :soccer:
- This PR is defined on top of PR #546, that refactors the model output API. The objective of this PR is to extend the Trainer class to support any T4Rec prediction tasks. 
- PR #546 should be merged to clean the code included in this PR.  
- Update `PyArrowDataloader` to support returning the targets. I also added a deprecation warning to remove  `PyArrowDataloader` in a future release as it is slow and not gpu-optimized. 
- Set the default type of inputs to `float32` (which is the default type expected by PyTorch).
- Squeeze the second dimension `1` of scalar inputs. 
- Add support for multiple predictions in the Trainer.
- Override the HF method `compute_loss` to pass the parameter `targets` to the model.
-  Remove HFWrapper module as it is not needed anymore with the custom `compute_loss` 

### Testing Details :mag:
- Update tests in `test_trainer.py`  to use merlin data loader instead of the `pyarrow` one,
- Add a unit-test to `test_trainer.py` that train/evaluate models with three different prediction tasks, using the Trainer class.


transformers4rec/torch/model/base.py
```diff
@@ -524,6 +521,17 @@ def forward(
         self, inputs: TensorOrTabularData, targets=None, training=False, testing=False, **kwargs
     ):
         # TODO: Optimize this
+        # Convert inputs to float32 which is the default type, expected by PyTorch
+        for name, val in inputs.items():
```
sararb, line 525: By default, PyTorch set the model's parameters as scalars (`float32`) and expects the input data to be of  type`float32`. 
If the inputs are Double (i.e. float64), the user should explicitly set the model to use Double parameters using `model.double()`. 
I am not sure if there is a use-case/benefit of using `Double` instead of scalar inputs? with the old default type convention, we were always using scalar inputs. That's why I added this conversion within the Model's forward method to always set the inputs as scalars.

transformers4rec/torch/trainer.py
```diff
@@ -106,7 +106,6 @@ def __init__(
     ):
 
         mock_dataset = DatasetMock()
-        hf_model = HFWrapper(model)
```
sararb, line 109: The wrapper is not needed as we are overriding all the HF Trainer methods that call the model's forward method: `prediction_step` and `compute_loss`

transformers4rec/torch/trainer.py
```diff
@@ -309,6 +308,31 @@ def get_scheduler(
...
+    def compute_loss(self, model, inputs, return_outputs=False):
+        """
+        Overriding :obj:`Trainer.compute_loss()`
+        To allow for passing the targets to the model's forward method
+        How the loss is computed by Trainer. By default, all Transformers4Rec models return
+        a dictionary of three elements {'loss', 'predictions', and 'labels}
+        """
+        inputs, targets = inputs
+        outputs = model(inputs, targets=targets, training=True)
+        # Save past state if it exists
```
sararb, line 320: This comment was copied from the HF original method `compute_loss` to track the #TODO if we upgrade the HF version.

transformers4rec/torch/utils/data_utils.py
```diff
@@ -72,6 +77,11 @@ def __init__(
             **kwargs,
         ):
             T4RecDataLoader.__init__(self)
+            warnings.warn(
```
sararb, line 80: Merlin data loader is now supporting CPU-based loading as well, so we might want to drop the `pyarrow` data loader in future release?

transformers4rec/torch/utils/data_utils.py
```diff
@@ -170,187 +182,180 @@ def from_schema(
...
-            dataset = validate_dataset(
-                self.paths_or_dataset,
-                self.batch_size,
-                buffer_size,
-                engine,
-                reader_kwargs,
+@dataloader_registry.register_with_multiple_names(
+    "merlin_dataloader", "merlin", "nvtabular_dataloader", "nvtabular"
+)
+class MerlinDataLoader(T4RecDataLoader, DLDataLoader):
```
sararb, line 212: I removed the condition `if dependencies.is_gpu_dataloader_available()` to always define the MerlinDataLoader (for cpu + gpu based environments)

gabrielspmoreira (approved):

PR #586, opened 2022-12-27
Small fixes in getting-started ETL and training notebooks and fix tuple error in serving notebook

rnyak:
This PR:

- fixes/sets data paths and env variables for certain args in the notebooks
- fixes the error due to returned tuple from Merlin dataloader in the `03` notebook. 
- adding a unit test with for `01` and `02` notebooks using testbook.

Note: if we want a unit test for serving notebook, we would need systems here. currently we dont couple TF4Rec to systems, that's why I removed that part in the unit test code. The decision is to do that in Merlin repo, instead of TF4Rec repo.


requirements/test.txt
```diff
@@ -4,3 +4,4 @@
 
 pytest>=5
 pytest-cov>=2
+testbook
```
edknv, line 7: It looks like the CPU test is failing because `tritonclient` is not installed. Perhaps we should add `tritonclient` here in `requirements/test.txt`?

rnyak: Thanks @edknv. Based on the discussion yesterday, I will remove the code script for serving nb from the unit test file so it should work now.

sararb (approved):
Thank you for the PR! The updates look good to me. I just left one comment regarding the old notebooks' unit test.

tests/unit/test_notebooks.py
```diff
@@ -28,7 +28,8 @@
 SESSION_PATH = "examples/getting-started-session-based/"
 
 
-@pytest.mark.skipif(find_spec("cudf") is None, reason="needs cudf")
+# @pytest.mark.skipif(find_spec("cudf") is None, reason="needs cudf")
+@pytest.mark.skip(reason="there is a testbook version of this test")
 def test_session(tmpdir):
```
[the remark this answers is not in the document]
rnyak: @jperez999 since I am replacing this test with the testbook version I added `mark.skip` into the existing test, but we can remove it as well. what do u think? thanks.

bschifferer:
The PR looks good to me. Do we have no Jenkins GitHub Action for it?

rnyak:
> Do we have no Jenkins GitHub Action for it?

Thanks for reviewing.  If I understood your question correctly : AFAIK, we do no longer use Jenkins for PR testing
. we only use GH actions.

jperez999 (approved):

jperez999:
it ran the new test and skipped the old one https://github.com/NVIDIA-Merlin/Transformers4Rec/actions/runs/3795493384/jobs/6454668622#step:3:326

PR #590, opened 2022-12-29
Add docstrings and the parameter to `row_groups_per_part ` to the MerlinDataLoader class

### LINKED ISSUE #550: [BUG] fix partition script in the DistributedDataParallel documentation
### Bug description

Currently in the document https://nvidia-merlin.github.io/Transformers4Rec/main/multi_gpu_train.html, we have the following script for users to repartiton the dataset. But this script is not working and is not repartitioning the dataset. 

`df.to_parquet("filename.parquet", row_group_size=10000)` 

This script should be modified in a way that the dataset is repartitioned properly.

sararb:
Fixes #550 

@bbozkaya runs different tests (see image below) of repartitioning a parquet file (using pandas or cudf) and it seems that `MerlinDataLoader` always loads the dataset files with 1 partition even though we partition to multiple groups when saving the parquet file (as recommended [here](https://github.com/NVIDIA-Merlin/Transformers4Rec/blob/9098a1d84823e9702578abeb15478e568d31703f/transformers4rec/torch/utils/data_utils.py#L256)).  To take into account these partitions, we should pass the parameter `row_groups_per_part=True`  to the merlin.io.Dataset.
image

### Goals :soccer:
- Add the parameter `row_groups_per_part` to `MerlinDataLoader` so as to load the dataset with the correct partitions.
- Add docstrings to the `MerlinDataLoader` to explain the different parameters. 
- Add a user warning to ensure that dataset's partitions are divisible by the number of GPUs for DDP training. This is needed to ensure optimal performance by equally distributing the data among available GPUs.


bbozkaya (approved):
I tested on 2 GPUs. It works fine when loading partitioned train data. For validation whoever, it seems to be using only 1 GPU and 1 partition. Is this expected or does it also need to be addressed?

sararb:
> I tested on 2 GPUs. It works fine when loading partitioned train data. For validation whoever, it seems to be using only 1 GPU and 1 partition. Is this expected or does it also need to be addressed?

Thank you for testing the solution! For the validation step, we rely on how HF transformers are setting the DDP training, and it seems that they don't wrap the model in a DDP mode for evaluation (training=False) ([here](https://github.com/huggingface/transformers/blob/a3c37825cc1e305dde63455b5f321586e6d29e07/src/transformers/trainer.py#L1358)). So it is expected that the validation runs on a single GPU but I don't know the motivation behind. I posted a question on HF forum to better understand the behavior of the Trainer in DDP+evaluation mode.

PR #600, opened 2023-01-17
add BC task script in readme

rnyak:


README.md
```diff
@@ -90,24 +90,87 @@ schema: tr.Schema = tr.data.tabular_sequence_testing_data.schema
...
+transformer_config = tr.XLNetConfig.build(
+    d_model=d_model, n_head=4, n_layer=2, total_seq_length=max_sequence_length
+)
+# Define the model block including: inputs, masking, projection and transformer block.
+body = tr.SequentialBlock(
+    input_module, tr.MLPBlock([64]), tr.TransformerBlock(transformer_config, masking=input_module.masking)
+)
+
+# Define a head with BinaryClassificationTask
+head = tr.Head(
```
mikemckiernan, line 158: My $0.02 is that the difference from the preceding example is too small to justify putting it here.

I've started updating the docstring for the BinaryClassificationTask to show this code sample.

Can you tell me what `tm` in the `metrics` represents?

mikemckiernan: I see it now, it's `import torchmetrics as tm`.

rnyak: `import torchmetrics as tm` yes. should I add it that line?

rnyak: @mikemckiernan `I've started updating the docstring for the BinaryClassificationTask to show this code sample.` does that mean you have a PR as well, therefore we dont need this PR?  Or where do you think we should add it this code snippet?

README.md
```diff
@@ -90,24 +90,87 @@ schema: tr.Schema = tr.data.tabular_sequence_testing_data.schema
...
+
+max_sequence_length, d_model = 20, 64
+
+# Define the input module to process the tabular input features.
+input_module = tr.TabularSequenceFeatures.from_schema(
+    schema,
+    max_sequence_length=max_sequence_length,
+    continuous_projection=d_model,
+    aggregation="concat",
+    masking="causal",
```
sararb, line 145: We could remove masking as in the general cases, the binary classification task is used with the full input sequence (without masking any information from the inputs)? 
```suggestion
    masking=None,
```

mikemckiernan:
I gave it a whirl--including some creative writing on the parameters to the class:
* Two sentences after the code sample in the README: https://mikemckiernan.github.io/Transformers4Rec/review/pr-6/README.html#sample-code-defining-and-training-the-model.  New edit, instead of "The preceding code..." I'd like to self edit  to "You can modify the preceding code..." so the text is user-focused.
* Put the sample code in an "Example usage" section in the [API desc](https://mikemckiernan.github.io/Transformers4Rec/review/pr-6/api/transformers4rec.torch.html#transformers4rec.torch.BinaryClassificationTask) of the class.

README.md
rnyak: @sararb thanks. will test it first, and change it to None.

rnyak:
> I gave it a whirl--including some creative writing on the parameters to the class:
> 
> * Two sentences after the code sample in the README: https://mikemckiernan.github.io/Transformers4Rec/review/pr-6/README.html#sample-code-defining-and-training-the-model.  New edit, instead of "The preceding code..." I'd like to self edit  to "You can modify the preceding code..." so the text is user-focused.
> * Put the sample code in an "Example usage" section in the [API desc](https://mikemckiernan.github.io/Transformers4Rec/review/pr-6/api/transformers4rec.torch.html#transformers4rec.torch.BinaryClassificationTask) of the class.
> 
> I drafted the change as a make-believe PR in my fork. If you like the changes, I can add the commit to this PR. It seemed too presumptuous to add a commit to your PR without checking.

@mikemckiernan it sounds good to me but we also change the masking to `None` in the input_module for BC task. so that'd be also different than the next-item prediction task.   I understand your point. you think the code snippets are very similar, and big chunk is repeated twice..   But I'd say better to repeat it than defining it insufficiently. what do you think?

mikemckiernan:
> 

I don't know if I'll persuade you to my thinking, but I'll try.  With so much similarity, the reader is using brain power to detect what is different (and wondering why two tasks are so overwhelmingly similar) than focusing on learning the concept--learning how to write the code to run a particular type of prediction task, classification, and so on.

I think it'd be better to include more of the sample code in the API reference section, as long as it is typical that _most_ folks will want to specify those blocks.  I'll make the change and we'll see if I understood correctly.  Thanks very much for your help and patience.

bschifferer (commented):
do we not need the schema.pb file in the examples?

transformers4rec/torch/model/prediction_task.py
```diff
@@ -44,6 +44,82 @@ def build(self, input_size) -> SequentialBlock:
...
+                summary_type="mean",
+                metrics=[
+                    tm.Precision(task='binary'),
+                    tm.Recall(task='binary'),
+                    tm.Accuracy(task='binary'),
+                    tm.F1Score(task='binary')
+                ]
+            ),
+            inputs=input_module,
+        )
```
bschifferer, line 89: It looks good - can we add the line until, we have the `model` object?

bschifferer (approved):

rnyak:
> do we not need the schema.pb file in the examples?

@bschifferer thanks for the comment. for the getting-started this schema file is not required, it is generated from NVT and read in then. We kept it there due to unit test, now we changed the unit test and it is run via testbook, we dont need this file either. This file creates confusion for the users so let's remove it :)

transformers4rec/torch/model/prediction_task.py
rnyak: thanks.

mikemckiernan: Added up to the model object.

PR #619, opened 2023-02-24
Start depending on Merlin models

marcromeyn:
### Goals :soccer:
This PR is meant to be a first small step into the direction of moving to the `Schema` definition in core.

This PR includes:
- Depending on merlin-models
- Copying `data/preprocess.py` to utils. This will make it possible to deprecated the data module in a future PR.


requirements/base.txt
```diff
@@ -1,6 +1,4 @@
-numpy>=1.17.0
 transformers<4.19
 tqdm>=4.27
-betterproto<2.0.0
 pyarrow>=1.0
-tensorflow-metadata
+merlin-models
```
nv-alaiacano, line 4: Should we provide a minimum version here just in case?

transformers4rec/utils/data_utils.py
```diff
@@ -0,0 +1,384 @@
+# type: ignore
```
nv-alaiacano, line 1: We should try to follow up pretty quickly with deprecating `preprocessing.py` so that we aren't stuck in a confusing state with two versions of this code at the same time. 

If it's not a ton of work (seems like it wouldn't be?) maybe tack that transition onto this PR?

marcromeyn: That's a good point. [This](https://github.com/NVIDIA-Merlin/Transformers4Rec/pull/619/commits/a2508999e5e8fa6ebc14b5d4f55da28e833c0504) commit does this.

setup.py
```diff
@@ -41,9 +41,9 @@ def read_requirements(filename):
 
 requirements = {
     "base": read_requirements("requirements/base.txt"),
+    "merlin": read_requirements("requirements/merlin.txt"),
```
oliverholworthy, line 44: why do we need an extra named `merlin` if this is going to be part of `install_requires`?

marcromeyn: This is to make it easier in CI/CD to install only the base-requirements & install all merlin-deps from the main branches.

oliverholworthy: Supporting the tox environment with installation from branches makes sense,

However, I wonder if we would like to restrict the keys we publish in `extras_requires` to only those that we expect users of the pacakge to use, since that becomes part of the interface to the package.

marcromeyn: Ah that's a good point. So that would imply changing it to: 

```python
"base": read_requirements("requirements/base.txt") + read_requirements("requirements/merlin.txt")
```

In that case maybe `base_external.txt` & `base_merlin.txt` might be better names?

oliverholworthy: yes I think that change would be clearer, both to setup.py and the rename of the requirements files

oliverholworthy (approved):

Issue #566, opened 2022-12-07
[FEA] Update Transformers4Rec to use merlin core schema 


sararb:
# 🚀 Feature request
Remove merlin-standardlib dependency and use the merlin-core schema instead.  

## Motivation
The merlin-standard lib has an old schema, that is not maintained. As the Merlin framework has now a `merlin-core` library that provides a Schema class and the related Tags. The latter schema is the one produced by Merlin NVTabular and also used by Merlin Systems for serving. So to maintain a seamless integration between T4Rec, NVTabular, and Systems, we should update T4Rec to use the same schema class used by NVTabular and Systems. 

## Scope
- [ ] Change notebooks to use merlin core schema
- [ ] Deprecation warnings

nv-alaiacano:
I see these 3 PRs related to this. Can we close this ticket?

* https://github.com/NVIDIA-Merlin/Transformers4Rec/pull/642
* https://github.com/NVIDIA-Merlin/Transformers4Rec/pull/634
* https://github.com/NVIDIA-Merlin/Transformers4Rec/pull/632

PR #621, opened 2023-02-27
move categorical_cardinalities out of the schema class

nv-alaiacano:
Continued small-piece-at-a-time refactoring for https://github.com/NVIDIA-Merlin/Transformers4Rec/issues/566

This removes `categorical_cardinalities` from the `Schema` class, which does not exist on the `merlin.schema.Schema` class.

### DIFF
--- a/merlin_standard_lib/__init__.py
+++ b/merlin_standard_lib/__init__.py
@@ -18,11 +18,19 @@ from betterproto import Message
 from merlin.schema.io import proto_utils
 
 from .schema import schema
-from .schema.schema import ColumnSchema, Schema
+from .schema.schema import ColumnSchema, Schema, categorical_cardinalities
 from .schema.tag import Tag
 
 # Other monkey-patching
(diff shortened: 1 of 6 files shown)


marcromeyn (changes requested):
As a general rule of thumb while removing schema-related functionality is worthwhile to check merlin-models. This because we have done this work before there to remove `merlin_standard_lib`. You can find `categorical_cardinalities` [here](https://github.com/NVIDIA-Merlin/models/blob/main/merlin/models/utils/schema_utils.py#L123), I think it's better to start using that function.

nv-alaiacano:
Oh, that's a great tip and will be helpful in the next step, which is to move from `merlin_standard_lib`'s `Tag` to `merlin.schema.Tags`.

Right now, this version of `categorical_cardinalities` still depends on the older version of `Tag` so the version in models isn't compatible. I'll be sure to make that change and depend on the one from models in the next PR, but for now I think this one needs to stay.

marcromeyn (approved):

PR #628, opened 2023-03-01
Cleanup shapes in model.input_schema and output_schema

rnyak:
This work addresses the Capturing shapes everywhere work. There are a bunch of places in Merlin that should fill information into the shape. for now we modified only model input_schema and output_schema code, but once we move to Schema from core, we can do more updates, accordingly.


transformers4rec/torch/model/base.py
```diff
@@ -738,7 +738,9 @@ def input_schema(self):
 
             dtype = {0: np.float32, 2: np.int64, 3: np.float32}[column.type]
             tags = column.tags
-            is_list = column.value_count.max > 0
+            dims = None
+            if column.value_count.max > 0:
+                dims = (None, column.value_count.max)
```
karlhigley, line 743: Do the `value_counts` always have the same size here? If so, it should still be safe to record the second dimension as `(column.value_count.min, column.value_count.max)` just in case they differ at some point (like after we add ragged input support to T4R.)

karlhigley (approved):

oliverholworthy (approved):

PR #631, opened 2023-03-07
Adjust serving notebook to account for underlying shape changes

karlhigley:
Those two changes are a workaround to make everything line up while we sort out shapes and ragged support in T4R, but should be okay for now AFAIK. We use padding in the NVT workflow, and capture the `sparse_max` list lengths in the schema, and the serving notebook then runs successfully.


examples/getting-started-session-based/03-serving-session-based-model-torch-backend.ipynb
```diff
@@ -70,8 +70,56 @@
      "text": [
```
rnyak, line 70: <pre class="ql-syntax" spellcheck="false">Line #1.    for col_name, col_schema in input_schema.column_schemas.items():
</pre><p>so this cell is needed to tell Ensemble api that model does not consume ragged inputs since value count min and max are same?  would <code>input_schema</code> wont print out required properties without executing this cell? </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/NVIDIA-Merlin/Transformers4Rec/pull/631/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='24'/>

rnyak:
> Those two changes are a workaround to make everything line up while we sort out shapes and ragged support in T4R, but should be okay for now AFAIK. We use padding in the NVT workflow, and capture the `sparse_max` list lengths in the schema, and the serving notebook then runs successfully.

thanks for that. LGTM. I only added one comment, out of curiosity. thanks!

examples/getting-started-session-based/03-serving-session-based-model-torch-backend.ipynb
karlhigley: Yeah, that's right. This is a manual step to capture the info from the `sparse_max` dictionary in the schemas, which helps the serving code know the shapes of the inputs we ended up with after padding in the Workflow.

karlhigley: In the longer term, we'll probably want to re-evaluate this and see if we can migrate from `sparse_max` to capturing shapes in the model schemas (which you're already working on the first part of.)

rnyak (approved):

PR #632, opened 2023-03-08
Migrate schema Tags to merlin.schema.Tags

nv-alaiacano:
Addresses https://github.com/NVIDIA-Merlin/Merlin/issues/844

Current status is that all tests pass except the getting-started notebooks.

Fixes # (issue)


tests/data/test_synthetic.py
```diff
@@ -31,5 +34,8 @@ def test_generate_item_interactions():
         "category": "int64",
         "item_recency": "float64",
     }
+    for key, val in dict(data.dtypes).items():
+        print((key, val))
```
nv-alaiacano, line 38: todo: remove

tests/data/test_synthetic.py
```diff
@@ -31,5 +34,8 @@ def test_generate_item_interactions():
...
+        assert val == expected_dtypes[key]
 
-    assert all(val == expected_dtypes[key] for key, val in dict(data.dtypes).items())
+    # assert all(val == expected_dtypes[key] for key, val in dict(data.dtypes).items())
```
nv-alaiacano, line 41: this definitely needs to go back in.

merlin_standard_lib/schema/schema.py
```diff
@@ -103,7 +102,7 @@ def create_continuous(
         tags: Optional[TagsType] = None,
         **kwargs,
     ) -> "ColumnSchema":
-        _tags: List[str] = [str(t) for t in tags] if tags else []
+        _tags: List[str] = [t.value for t in TagSet(tags or [])]
```
nv-alaiacano, line 105: This is something I don't love - `_tags` is a `List[str]` not a `TagsType`. We convert actual `Tags` to their string value, but I don't think we convert them back anywhere. I need to understand it a little more now that the tests are passing (hopefully)

marcromeyn: I agree. I think we converted to strings everywhere to make it easier for users to provide custom tags as strings. Not completely sure how we handle that currently across Merlin.

marcromeyn (approved):

PR #634, opened 2023-03-09
Allow Schema class from core to be used to create input-blocks

marcromeyn:
### Goals :soccer:
The goal is to deprecate the `Schema` class in `merlin_standard_lib` in favor of the one in merlin-core. In order to give our users some time for this breaking change, this PR adds support for providing both the Schema definitions to create input-blocks. 

The next step would be to do the same for the `Trainer` class.

### Testing Details :mag:
This PR adds some parameterization in order to test the creation of input-blocks from both the Schema definitions.

### DIFF
--- a/transformers4rec/torch/tabular/base.py
+++ b/transformers4rec/torch/tabular/base.py
@@ -15,6 +15,7 @@
 #
 
 from abc import ABC
+from copy import deepcopy
 from functools import reduce
 from typing import Dict, List, Optional, Tuple, Union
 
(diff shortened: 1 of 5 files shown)


transformers4rec/torch/tabular/base.py
```diff
@@ -186,7 +187,7 @@ def from_schema(cls, schema: Schema, tags=None, **kwargs) -> Optional["TabularMo
         -------
         Optional[TabularModule]
         """
-        schema_copy = schema.copy()
+        schema_copy = deepcopy(schema)
```
oliverholworthy, line 190: is this `deepcopy` required for the Merlin Core Schema?

oliverholworthy: We could add a copy method to the core schema that would avoid the need for `deepcopy`

https://github.com/NVIDIA-Merlin/core/pull/240

oliverholworthy (approved):

PR #638, opened 2023-03-10
Allow Schema class from core to be used to create TabularSequenceFeatures

marcromeyn:
### Goals :soccer:
Follow up to #634, this PR allows the `Schema` class from merlin-core to instantiate `TabularSequenceFeatures`.

### Testing Details :mag:
This PR also includes a change to make `parameterize_schema` more generic by making the schema name an argument.

### DIFF
--- a/transformers4rec/torch/features/embedding.py
+++ b/transformers4rec/torch/features/embedding.py
@@ -161,6 +161,11 @@ class EmbeddingFeatures(InputBlock):
 
         _item_id = schema.select_by_tag(Tags.ITEM_ID)
         if not item_id and len(_item_id) > 0:
+            if len(_item_id) > 1:
+                raise ValueError(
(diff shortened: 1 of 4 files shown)


transformers4rec/torch/features/embedding.py
```diff
@@ -159,8 +159,9 @@ def from_schema(  # type: ignore
         if tags:
             schema = schema.select_by_tag(tags)
 
-        if not item_id and schema.select_by_tag(["item_id"]).column_names:
-            item_id = schema.select_by_tag(["item_id"]).column_names[0]
+        _item_id = schema.select_by_tag(Tags.ITEM_ID)
+        if not item_id and len(_item_id) > 0:
```
oliverholworthy, line 163: if for some reason the select_by_tag matches more than one column, maybe would like to raise an exception?

oliverholworthy (approved):

PR #641, opened 2023-03-10
Fix the predictions returned by `Trainer.predict(..)`

### LINKED ISSUE #636: [BUG] BinaryClassificationTask prediction does not return one score per session with summary_type='mean'
### Bug description

When we train a BinaryClassificationTask and then do 

```
prediction = trainer.predict(eval_paths[0])
prediction
```
this does not return prediction scores. 

In addition, we are able to set  `trainer.args.predict_top_k`, but we should not be able to do that for BC Task, since this is particularly designed for NextItem Prediction task.

### Expected behavior

we should be getting a probability score per session.

### Environment details
- Transformers4Rec version: 
- Platform:
- Python version:
- Huggingface Transformers version:
- PyTorch version (GPU?):
- Tensorflow version (GPU?):

### Additional context
using `merlin-pytorch:22.12` with the latest main branches pulled.

sararb:
Fixes #636

### Goals :soccer:
- Fix the predictions returned by `Trainer.predict()`.
- Ensure `predict_top_k` can only be applied to NextItemPredictionTask. 
- Fix the output shapes of the model's predictions in the sequential binary classification or the regression task.

### Testing Details :mag:
- Add `test_sequential_binary_classification_model` to test a sequential model trained with a per-item binary classification task. 
- Checks the shapes of predictions in `test_trainer_music_streaming`
- Add `test_trainer_trop_k_with_wrong_task` to ensure the BC task cannot be used with `predict_top_k > 0`

### DIFF
--- a/transformers4rec/torch/model/base.py
+++ b/transformers4rec/torch/model/base.py
@@ -36,7 +36,7 @@ from merlin_standard_lib import Schema
 from ..block.base import BlockBase, BlockOrModule, BlockType
 from ..features.base import InputBlock
 from ..features.sequence import TabularFeaturesType
-from ..typing import TabularData, TensorOrTabularData
+from ..typing import TabularData
 from ..utils.torch_utils import LossMixin, MetricsMixin
 
 
@@ -517,9 +517,7 @@ class Model(torch.nn.Module, LossMixin, MetricsMixin):
         self.head_reduction = head_reduction
         self.optimizer = optimizer
 
-    def forward(
-        self, inputs: TensorOrTabularData, targets=None, training=False, testing=False, **kwargs
-    ):
+    def forward(self, inputs: TabularData, targets=None, training=False, testing=False, **kwargs):
         # Convert inputs to float32 which is the default type, expected by PyTorch
         for name, val in inputs.items():
             if torch.is_floating_point(val):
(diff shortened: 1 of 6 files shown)


transformers4rec/torch/model/base.py
```diff
@@ -517,9 +517,7 @@ def __init__(
         self.head_reduction = head_reduction
         self.optimizer = optimizer
 
-    def forward(
-        self, inputs: TensorOrTabularData, targets=None, training=False, testing=False, **kwargs
-    ):
+    def forward(self, inputs: TabularData, targets=None, training=False, testing=False, **kwargs):
```
sararb, line 520: T4Rec only supports a dictionary of tensors (i.e. TabularData) as inputs.

transformers4rec/torch/model/prediction_task.py
```diff
@@ -36,7 +36,7 @@ def build(self, input_size) -> SequentialBlock:
         return SequentialBlock(
             torch.nn.Linear(input_size[-1], 1, bias=False),
             torch.nn.Sigmoid(),
-            LambdaModule(lambda x: x.view(-1)),
+            LambdaModule(lambda x: torch.squeeze(x, -1)),
```
sararb, line 39: In the case of sequential binary classification, we'll need to keep the second dimension (i.e. sequence length)

rnyak:
rerun tests

rnyak (approved):

PR #643, opened 2023-03-13
fix test_remove_consecutive_interactions

rnyak:


tests/utils/test_data_utils.py
```diff
@@ -13,8 +13,7 @@ def test_remove_consecutive_interactions(testing_data):
         df.copy(), timestamp_col="event_timestamp"
     )
 
-    assert len(filtered) < len(df)
```
sararb, line 16: We still need to verify whether the `remove_consecutive_interactions` method is functioning as expected ( i.e. whether it is dropping repeated interactions).

For that, we could create duplicated interactions and add them to the dataframe `df`. Something like: 
```
    
    #generate repeated interactions
    repeated = 3
    rows = []
    for _ in range(repeated):
        new_row = [123, 42, 20, 10, 0.23, 0.1, 0.4, 0.23, 0.41, [2, 3], 1, 1]
        rows.append(new_row)
    to_add = type(df)(rows, columns=df.columns)
    
    df = df.append(to_add, ignore_index=True)
 ```

sararb (commented):
@rnyak, Thank you for the PR! I left one suggestion to test whether the repeated interactions are correctly dropped.

tests/utils/test_data_utils.py
rnyak: thanks. added these lines to the test script.

sararb (approved):

rnyak:
rerun tests

PR #647, opened 2023-03-14
Update padding of ragged features to enable dataloader change 

oliverholworthy:
### Goals :soccer:

- Enable a change to the merlin dataloader (https://github.com/NVIDIA-Merlin/dataloader/pull/103) that removes sparse output type and padding based on the value count property.
- Implement as a backwards compatible change for the current version as well as the future version  https://github.com/NVIDIA-Merlin/dataloader/pull/101

- Implements the equivalent of the padding currently in the dataloader
  - Uses ragged -> sparse -> dense conversion which appears to be faster than an alternative approach assigning values to a tensor constructed with zeros 

### Testing Details :mag:

- Existing tests should cover existing usage of the Merlin Dataloader
- Adds unit tests for padding function to check pad batch

### DIFF
--- a/transformers4rec/torch/utils/data_utils.py
+++ b/transformers4rec/torch/utils/data_utils.py
@@ -356,7 +356,7 @@ class MerlinDataLoader(T4RecDataLoader, DLDataLoader):
             global_size=global_size,
             global_rank=global_rank,
             drop_last=drop_last,
-        ).map(self.get_pad_fn(sparse_max))
+        ).map(self._get_pad_fn(sparse_max))
 
         DLDataLoader.__init__(
             self,
--- a/transformers4rec/torch/utils/data_utils.py
+++ b/transformers4rec/torch/utils/data_utils.py
@@ -380,9 +380,9 @@ class MerlinDataLoader(T4RecDataLoader, DLDataLoader):
     @staticmethod
     def _augment_schema(
         schema,
-        cats=None,
-        conts=None,
-        labels=None,
+        cats,
+        conts,
+        labels,
     ):
         schema = schema.select_by_name(conts + cats + labels)
 
--- a/transformers4rec/torch/utils/data_utils.py
+++ b/transformers4rec/torch/utils/data_utils.py
@@ -28,7 +28,6 @@ from torch.utils.data import DataLoader as PyTorchDataLoader
 from torch.utils.data import Dataset, IterableDataset
 
 from merlin_standard_lib import Schema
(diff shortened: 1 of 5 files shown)


transformers4rec/torch/utils/data_utils.py
```diff
@@ -367,6 +368,34 @@ def __init__(
...
+        return pad_fn
+
+    @staticmethod
+    def _augment_schema(
+        schema,
+        cats=None,
+        conts=None,
+        labels=None,
+    ):
+        schema = schema.select_by_name(conts + cats + labels)
```
marcromeyn, line 387: Maybe we should remove the `=None` for cats, conts & labels?

transformers4rec/torch/utils/padding.py
```diff
@@ -0,0 +1,118 @@
...
+    ------
+    ValueError
+        If ragged column found with no padding length provided
+    """
+    if batch is None or not isinstance(batch, dict):
+        return batch
+
+    batch_padded = {}
+    for k, values in batch.items():
+        if k.endswith("__values"):
```
marcromeyn, line 84: Wouldn't it be better to put `__values` and `__offsets` as constants somewhere in merlin-core?

transformers4rec/torch/utils/data_utils.py
```diff
@@ -367,6 +368,34 @@ def __init__(
...
+    def _get_pad_fn(padding_lengths):
+        def pad_fn(x, y):
+            new_x = pad_batch(x, padding_lengths)
+            new_y = pad_batch(y, padding_lengths)
+            return new_x, new_y
+
+        return pad_fn
+
+    @staticmethod
+    def _augment_schema(
```
oliverholworthy, line 381: Moved this from `schema_utiils` to here so that it's closer to the only place it's called in the codebase.

transformers4rec/torch/utils/data_utils.py
oliverholworthy: that's clearer without the default now. it wouldn't have worked as None before this change either.

oliverholworthy: On second thought, I've put the None back and added support for the None to the method by setting to an empty list if not provided. 

It doesn't appear to be captured by any tests, but this would enable training a model with only continuous or only categorical features. Currently you need to have at least one of each for the current version of this MerlinDataloader to work.

karlhigley (approved):

PR #648, opened 2023-03-15
Print deprecation warning for usage of Schema from merlin_standard_lib

marcromeyn:
Fixes # (issue)

### DIFF
--- a/merlin_standard_lib/schema/schema.py
+++ b/merlin_standard_lib/schema/schema.py
@@ -16,6 +16,7 @@ import collections
 import json
 import os
 from typing import Any, Callable, Dict, List, Optional, Sequence, Tuple, TypeVar, Union
+from warnings import warn
 
 from google.protobuf import json_format, text_format
 from google.protobuf.message import Message as ProtoMessage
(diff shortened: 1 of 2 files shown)


merlin_standard_lib/schema/schema.py
```diff
@@ -216,6 +217,18 @@ class Schema(_Schema):
     """A collection of column schemas for a dataset."""
 
     feature: List["ColumnSchema"] = betterproto.message_field(1)
+    _is_first_init = True
+
+    def __post_init__(self):
+        super().__post_init__()
+        if self._is_first_init:
+            # TODO: Make description more descriptive. What version are we planning to remove it?
```
marcromeyn, line 225: @nv-alaiacano @oliverholworthy How should we write this warning? Do we want to commit to a specific release to actually remove it?

nv-alaiacano (approved):

JiwaniZakir (commented):
The `warn()` call in `schema.py` has a bug: it passes two separate string arguments rather than one concatenated message. `warnings.warn(message, category, stacklevel)` treats the second positional argument as the warning category, so `"use Schema from \`merlin.schema\` instead."` is being passed as `category` (a `str`, not a `Warning` subclass), and `DeprecationWarning` ends up as the `stacklevel` argument. This will either raise a `TypeError` or silently misbehave at runtime. The fix is to concatenate the strings: `warn("Schema from \`merlin_standard_lib\` is deprecated, use Schema from \`merlin.schema\` instead.", DeprecationWarning, stacklevel=2)`. Note that `stacklevel=2` is also preferable so the warning points to the caller's code rather than to `__post_init__` itself.

The `_is_first_init` class variable approach is also fragile — it suppresses the warning for all subsequent instantiations process-wide, and the test has to manually reset it (`schema.Schema._is_first_init = True`) to get a warning to fire, which is a sign the design is working against testability. A more idiomatic approach would be to use `warnings.warn` with `stacklevel=2` and rely on Python's built-in per-location deduplication of `DeprecationWarning`, which already suppresses repeated warnings from the same call site without requiring a custom flag.

PR #649, opened 2023-03-15
Update minimum version of merlin-models to 23.4.0

oliverholworthy:
Transformers4Rec now requires merlin-models 23.4.0 (upcoming release) to run correctly.

Depends on https://github.com/NVIDIA-Merlin/models/pull/1069


requirements/base_merlin.txt
```diff
@@ -1 +1 @@
-merlin-models>=0.11
+merlin-models>=23.3.0
```
rnyak, line 1: shall we update it to 23.4.0?

oliverholworthy: Yes, we'll need to update this to 23.4.0 once the models package is published

oliverholworthy: updated

oliverholworthy:
> @oliver is that `merlin-models>=23.4` supposed to be `23.04` or does not matter?

`23.4.0` is the version that PyPI receives. The [leading zeros are stripped out by setuptools](https://setuptools.pypa.io/en/latest/userguide/distribution.html). The version specifier in the requirements will work with leading zeros too

rnyak (approved):

PR #642, opened 2023-03-13
Allow Schema class from core to be used to create Trainer

marcromeyn:
Fixes # (issue)

PR #650, opened 2023-03-15
Update example notebooks to create schema object from merlin core

rnyak:
This PR updates schema creation line in the example notebooks based on this merged PR https://github.com/NVIDIA-Merlin/Transformers4Rec/pull/642.

Blocked by the error coming from multi gpu training example: https://github.com/NVIDIA-Merlin/Transformers4Rec/issues/651


examples/end-to-end-session-based/01-ETL-with-NVTabular.ipynb
```diff
@@ -69,7 +69,18 @@
    "execution_count": 2,
```
bschifferer, line 69: <p>I think we call fit transform twice - first above - .fit and .transform and then .compute</p><p> </p><p>and here, we call workflow.fir_transform().to_parquet.</p><p> </p><p>We do the calculation twice.</p><p> </p><p>I think we should change the process above to all fit_transform().to_parquet and then read the data with cudf.read_parquet() here.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/NVIDIA-Merlin/Transformers4Rec/pull/650/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='35'/>

examples/getting-started-session-based/01-ETL-with-NVTabular.ipynb
```diff
@@ -1,1059 +1,932 @@
 {
```
bschifferer, line 1: <p>Same - we compute the workflow twice (fit transform) - would it not better to store it to disk and then read it to split it into days?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/NVIDIA-Merlin/Transformers4Rec/pull/650/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='20'/>

PR #663, opened 2023-04-04
Fix bug in get_output_sizes_from_schema with core-schema

marcromeyn:
### Goals :soccer:
Fixes bug that was reported by @rnyak.

### Testing Details :mag:
Added a unit-test ensuring that the fix works.


oliverholworthy (approved):
This looks ok. Did this fix the issue with context features and the core schema or are there further changes required?

PR #666, opened 2023-04-05
Add support for ragged inputs to model

oliverholworthy:
Part of https://github.com/NVIDIA-Merlin/Merlin/issues/255

### Goals :soccer:

- Enable Transformers4Rec model to be called with ragged input representation.

- Adds pre-processing step to the first part of the forward method of the model that pads any tensors in the ragged representation.
  - Where there are two tensors with names `{feature}__values` `{feature}__offests`.
  - Pads all to minimum of the maximum sequence in batch or the model max_sequence_length (if defined)

### Testing Details :mag:

- Adds a test for model with sequence inputs and passing ragged representation inputs

### DIFF
--- a/tests/unit/utils/test_padding.py
+++ b/tests/unit/utils/test_padding.py
@@ -17,7 +17,7 @@ from itertools import accumulate
 
 import torch
 
-from transformers4rec.torch.utils.padding import pad_batch
+from transformers4rec.torch.utils.padding import pad_batch, pad_inputs
 
 
 def _get_values_offsets(data):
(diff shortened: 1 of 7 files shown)


transformers4rec/torch/model/base.py
```diff
@@ -511,13 +516,50 @@ def __init__(
...
+
+        padding_sequence_length = batch_max_sequence_length
+        if self.max_sequence_length is not None:
+            padding_sequence_length = min(self.max_sequence_length, batch_max_sequence_length)
+
+        if padding_sequence_length:
+            padding_lengths = {}
+            for name in inputs.keys():
+                if name.endswith("__offsets"):
+                    padding_lengths[name[:-9]] = padding_sequence_length
```
marcromeyn, line 548:
```suggestion
                    padding_lengths[name[:-len("__offsets")]] = padding_sequence_length
```

transformers4rec/torch/utils/padding.py
```diff
@@ -77,15 +77,15 @@ def pad_batch(batch: Batch, padding_lengths: Dict[str, int]) -> Batch:
     ValueError
         If ragged column found with no padding length provided
     """
-    if batch is None or not isinstance(batch, dict):
-        return batch
-
     batch_padded = {}
-    for col_name, col in TensorTable(batch).items():
```
oliverholworthy, line 84: TensorTable is not currently compatible with `torch.jit.script` compilation. 

example of one of the errors that shows up (I don't think it prints out all errors, only the first it encounters -> there may be more unsupported things apart from the below example)

```
E   torch.jit.frontend.UnsupportedNodeError: SetComp aren't supported:
E     File "/workspace/merlin/core/merlin/table/tensor_table.py", line 61
E       def _validate_columns(self, cols_dict):
E           col_types = {type(col_obj) for col_obj in cols_dict.values()}
E                       ~ <--- HERE
E           if len(col_types) >= 2:
E               raise TypeError(
E   '__torch__.merlin.table.tensor_table.TensorTable' is being compiled since it was called from 'pad_batch'
```

tests/unit/utils/test_padding.py
```diff
@@ -30,68 +30,122 @@ def _get_values_offsets(data):
...
+        assert torch.equal(
+            padded_inputs["a"],
+            torch.tensor(
+                [
+                    [1, 2, 3],
+                    [6, 7, 8],
+                ]
+            ),
+        )
+        assert torch.equal(
```
oliverholworthy, line 148: dense sequence inputs are not padded as part of this `pad_inputs` currently. Assuming we'll either have ragged or padded sequence inputs, not a mix of both

transformers4rec/torch/model/base.py
```diff
@@ -481,6 +482,7 @@ def __init__(
         head_reduction: str = "mean",
         optimizer: Type[torch.optim.Optimizer] = torch.optim.Adam,
         name: str = None,
+        max_sequence_length: Optional[int] = None,
```
oliverholworthy, line 485: Added a `max_sequence_length` to limit the size of the padding when receiving ragged inputs.

transformers4rec/torch/utils/padding.py
```diff
@@ -95,8 +95,56 @@ def pad_batch(batch: Batch, padding_lengths: Dict[str, int]) -> Batch:
...
+            if padding_length is not None:
+                batch_padded[col_name] = _pad_dense_tensor(value, padding_length)
+            else:
+                batch_padded[col_name] = value
 
     return batch_padded
+
+
+@torch.jit.script
+def pad_inputs(inputs: Dict[str, torch.Tensor], max_sequence_length: Optional[int] = None):
```
marcromeyn, line 112:
```suggestion
def pad_inputs(
    inputs: Dict[str, torch.Tensor], max_sequence_length: Optional[int] = None
) -> Dict[str, torch.Tensor]:
```

marcromeyn (approved):

PR #678, opened 2023-04-18
Add topk arg to return topk items and scores at inference step

rnyak:
This PR adds functionality for returning topk most relevant (with the highest scores) item ids from Triton IS, for NextItemPrediction task.

Current blocker:

~~The code designed to return top_k item ids (int 64 dtype), but model.output_schema returns `next_item` as float32 dtype, which creates an error from Triton.~~

Shall we change the code base in a way that model.output_schema matches with the expected output and output dtype from Triton? Or shall we return top_k item id scores, instead of item_ids?

**Status update:**

After modifying the model.output_schema, we can now return two outputs (item_scores, item_ids) from Triton.

Remaining tasks:
- [x] be sure the dtype of categorical item-id in the model.output schema matches with the model.input_schema
- [x] add a unit test
- [x] add an example notebook to showcase topK layer or modify one of the existing example: will be taken care of by this PR https://github.com/NVIDIA-Merlin/Transformers4Rec/pull/680

### DIFF
--- a/transformers4rec/torch/model/base.py
+++ b/transformers4rec/torch/model/base.py
@@ -371,7 +371,7 @@ class Head(torch.nn.Module, LossMixin, MetricsMixin):
         testing: bool = False,
         targets: Union[torch.Tensor, TabularData] = None,
         call_body: bool = False,
-        top_k: Optional[int] = -1,
+        top_k: Optional[int] = None,
         **kwargs,
     ) -> Union[torch.Tensor, TabularData]:
         outputs = {}
@@ -497,7 +497,7 @@ class Model(torch.nn.Module, LossMixin, MetricsMixin):
         optimizer: Type[torch.optim.Optimizer] = torch.optim.Adam,
         name: str = None,
         max_sequence_length: Optional[int] = None,
-        top_k: Optional[int] = -1,
+        top_k: Optional[int] = None,
     ):
         """Model class that can aggregate one or multiple heads.
         Parameters
@@ -517,7 +517,7 @@ class Model(torch.nn.Module, LossMixin, MetricsMixin):
             Used to truncate sequence inputs longer than this value.
         top_k: int, optional
             The number of items to return at the inference step once the model is deployed.
-            Default is -1, which will return all items.
+            Default is None, which will return all items.
         """
         if head_weights:
             if not isinstance(head_weights, list):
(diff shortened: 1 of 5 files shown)


transformers4rec/torch/model/base.py
```diff
@@ -400,9 +403,20 @@ def forward(
             outputs = {"loss": loss, "labels": labels, "predictions": predictions}
         else:
             for name, task in self.prediction_task_dict.items():
-                outputs[name] = task(
-                    body_outputs, targets=targets, training=training, testing=testing, **kwargs
-                )
+                if isinstance(task, NextItemPredictionTask):
```
oliverholworthy, line 406: If the `NextItemPredictionTask` is always something that is created by user code. An alternative to passing to the foward method could be accepting `top_k` in the constructor `__init__` method. That way we wouldn't need to have this condition here

transformers4rec/torch/model/prediction_task.py
```diff
@@ -342,7 +344,11 @@ def forward(self, inputs: torch.Tensor, targets=None, training=False, testing=Fa
             # Compute predictions probs
             x, _ = self.pre(x)  # type: ignore
 
-            return x
+            if top_k == -1:
```
oliverholworthy, line 347: Does top_k need to be a numeric value? Would it work with with `None` as the default value, for example?

transformers4rec/torch/model/base.py
rnyak: The reason we did not add this to the constructor is bcs there is no turning back changing the topK value at the inf step, once the model is trained with top_k arg. So to make it flexible, we decided to add it in the forward method.
```
head = tr.Head(
    body,
    tr.NextItemPredictionTask(weight_tying=True, 
                              metrics=metrics, top_k =20),
    inputs=inputs,
)
```

@sararb wanna add something else?

transformers4rec/torch/model/prediction_task.py
rnyak: I think `-1` is a used convention but we can set it to None either. @sararb will setting it to None create any issue?

sararb: You're right, it is just a convention. Setting it to None won't create any issue.

sararb (approved):

oliverholworthy (approved):

oliverholworthy (approved):

rnyak:
rerun tests

rnyak:
rerun tests

PR #690, opened 2023-05-02
[DRAFT] Add pre-trained embeddings support to T4Rec input blocks 

sararb:
Fixes  #683 #684 #685 #485

### Goals :soccer:
- [x]  Implement a `PretrainedEmbeddingFeature` that processes pre-trained input features. #485
- [x]  Update TabularFeatures to support input from pretrained embeddings, with optional projection Transformers4Rec. #683
- [x]  Update TabularSequenceFeatures to support input from pretrained embeddings, with optional projection Transformers4Rec #684 
- [x] Convert pre-trained embeddings into 2D/3D dense tensor representation Transformers4Rec #685
- [ ] Add pre-trained embeddings support to the T4Rec data loader class `MerlinDataLoader` (which is a wrapper of the merlin torch loader class) 
- [ ] Broadcast 2-D pre-trained embeddings to be used with sequential input block: TabularSequenceFeatures 
- [ ] Aggregate 3-D pre-rained embeddings to be used with non-sequential input block: TabularFeatures 
- [ ] Add LayerNorm option to PretrainedEmbeddingFeature
- [x] Implement unit test of Merlin data loader with pre-trained features + T4Rec input blocks 
- [ ] Implement an end-to-end test using Trainer + merlin loader with pre-trained transform operator

PR #696, opened 2023-05-10
[DRAFT] Remove the custom padding in T4Rec's MerlinDataLoader

sararb:
### Goals :soccer:
The Merlin data loader has a Padding operator that we can pass via the transforms parameter. To avoid code redundancy, I created a PR in T4Rec that removes custom padding from the T4Rec's `MerlinDataLoader` class. 
In this PR, the default behavior of the Trainer is to set the data loader without padding. If the user wants to perform padding outside of the model, they can explicitly provide a data loader with a Padding transform.

- Removes the `sparse_max` logic from `MerlinDataLoader`
- Extend `MerlinDataLoader` to accept the `transforms` parameter where the user can pass the Padding op. 

### Testing Details :mag:
- Add a unit test to check that we can provide a data loader with Padding op to the Trainer class

### DIFF
--- a/tests/unit/torch/test_trainer.py
+++ b/tests/unit/torch/test_trainer.py
@@ -600,7 +600,7 @@ def test_trainer_trop_k_with_wrong_task():
     assert "Top-k prediction is specific to NextItemPredictionTask" in str(excinfo.value)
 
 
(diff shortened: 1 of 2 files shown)


sararb:
Closing this as a solution in favor of updating the T4Rec padding code is implemented in #690

PR #712, opened 2023-06-01
Remove unused features from sequence testing data

oliverholworthy:
### Goals :soccer:

Update example dataset to only include features that can be proceseed by the model. To simplify our examples.

Removes features from the sequece testing dataset schema that don't have  a continuous or categorial tag.

### Testing Details :mag:

Existing tests


sararb (approved):
Thank you for the PR, it looks good to me!

PR #716, opened 2023-06-07
Remove deps from integration test env

oliverholworthy:
Remove deps from integration test env

Checking if this makes any difference to the CUDA version error we're  seeing


oliverholworthy:
Test still fails with the same error without the 'cu11' packages installed (https://github.com/NVIDIA-Merlin/Transformers4Rec/actions/runs/5204937605/jobs/9389802393?pr=716)

PR #717, opened 2023-06-07
Extend LD_LIBRARY_PATH with pytorch directory

oliverholworthy:
Follow-up to #715 Extends LD_LIBRARY_PATH with pytorch directory instead of overwriting it.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -55,7 +55,6 @@ commands =
 sitepackages=true
 setenv = 
     TF_GPU_ALLOCATOR=cuda_malloc_async
-    LD_LIBRARY_PATH=/opt/tritonserver/backends/pytorch
 deps =
     -rrequirements/test.txt   
 commands =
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -66,7 +65,7 @@ commands =
     python -m pip install --upgrade git+https://github.com/NVIDIA-Merlin/systems.git@{posargs:main}
     python -m pip install --upgrade git+https://github.com/NVIDIA-Merlin/models.git@{posargs:main}
 
-    python -m pytest -rsx --cov-config tests/.coveragerc --cov-report term-missing --cov=. tests/integration
+    LD_LIBRARY_PATH="/opt/tritonserver/backends/pytorch:$LD_LIBRARY_PATH" python -m pytest -rsx --cov-config tests/.coveragerc --cov-report term-missing --cov=. tests/integration
```
nv-alaiacano, line 68: is it required to add `LD_LIBRARY_PATH` to tox's `passenv`?

oliverholworthy: That's not required  because `LD_LIBRARY_PATH` is one of the few that is passed through by default along with other's like `PATH`.

I've updated this PR to keep the logic in `setenv` now

nv-alaiacano (approved):

oliverholworthy:
This appears to get the test working. One thing that's still a bit unclear to me is that this test runs ok outside the tox environment with LD_LIBRARY_PATH replaced entirely with `/opt/tritonserver/backends/pytorch`. My guess is that there  are other environment variables that help figure out where to find the correct cuda versions to use that make the rest of LD_LIBRARY_PATH not strictly required

PR #723, opened 2023-06-21
Fix the CLM performance mismatch between evaluation and manual inference

### LINKED ISSUE #719: [BUG] XLNET-CLM eval recall metric value does not match with custom np based recall metric value
### Bug description
When we train an XLNet model with `CLM` masking, the model prints out its own evaluation metrics (ndcg@k, recall@k, etc.) from `trainer.evaluate()` step. If we want to apply our own custom metric func using numpy something like below, the metric values do not match, but they match if we use `MLM` masking instead.

```
def recall(predicted_items: np.ndarray, real_items: np.ndarray) -> float:
    bs, top_k = predicted_items.shape
    valid_rows = real_items != 0

    # reshape predictions and labels to compare
    # the top-10 predicted item-ids with the label id.
    real_items = real_items.reshape(bs, 1, -1)
    predicted_items = predicted_items.reshape(bs, 1, top_k)

    num_relevant = real_items.shape[-1]
    predicted_correct_sum = (predicted_items == real_items).sum(-1)
    predicted_correct_sum = predicted_correct_sum[valid_rows]
    recall_per_row = predicted_correct_sum / num_relevant
    return np.mean(recall_per_row)
```

coming soon. 

### Environment details
- Transformers4Rec version: 
- Platform:
- Python version:
- Huggingface Transformers version:
- PyTorch version (GPU?):
- Tensorflow version (GPU?):

sararb:
Fixes #719 

### Goals :soccer:
The current CLM strategy is as follows 
image
The main issues of the current implementation are: 
- We mask useful past information when we evaluate on the last item only ==> This is wrong as past information is a valuable context for predictions. 
- At inference, the padded positions are represented with `0`-embeddings while during training these positions are replaced with trainable [MASK] embeddings. ==> We should have the same training, evaluation, and inference representation strategy. 

- Updated the class `CausalLanguageModeling` to: 
    -  Replace padded positions with [MASK] embeddings at inference. 
    - During the evaluation of the last item, I defined the `label_mask` as the padding mask information (to keep information about actual past items). 

- I ran the `t4r_paper_repro` script using 5 days of ecomrees46 dataset and these are the results: 
```
**CLM run using main branch (w/o fix)**
Recall@10 of manually masked test data = 0.3915199603272998
eval//next-item/recall_at_10 0.25108

**Fix the evaluation on last item by not masking the past information**
eval_/next-item/recall_at_10 = 0.2980385422706604
Recall@10 of manually masked test data = 0.29812381188527975
```

### Testing Details :mag:
- Changed `test_mask_only_last_item_for_eval` to get target mask information from `lm.masked_targets` 
- Changed `test_sequential_tabular_features_ignore_masking` as the inference mode of CLM is changing the inputs by replacing `0` padded positions with  [MASK] embeddings

## Future work 
- The [MASK] embeddings is not used to generate the next-item prediction scores and this raises a question about whether we should remove it and just use `0`-embedding to represent padded positions. 
==> We need to re-run T4Rec paper experiments without [MASK] variable and check how the evaluation results are impacted.

### DIFF
--- a/transformers4rec/torch/masking.py
+++ b/transformers4rec/torch/masking.py
@@ -274,6 +274,10 @@ class CausalLanguageModeling(MaskSequence):
     def _compute_masked_targets(
         self, item_ids: torch.Tensor, training: bool = False, testing: bool = False
     ) -> MaskingInfo:
+        if not training and not testing:
+            mask_labels = item_ids != self.padding_idx
+            return MaskingInfo(mask_labels, item_ids)
+
         masking_info = self.predict_all(item_ids)
         mask_labels, labels = masking_info.schema, masking_info.targets
 
@@ -290,7 +294,8 @@ class CausalLanguageModeling(MaskSequence):
             label_seq_trg_eval[rows_ids, last_item_sessions] = labels[rows_ids, last_item_sessions]
             # Updating labels and mask
             labels = label_seq_trg_eval
-            mask_labels = label_seq_trg_eval != self.padding_idx
+            # We only mask padded positions
+            mask_labels = item_ids != self.padding_idx
 
         return MaskingInfo(mask_labels, labels)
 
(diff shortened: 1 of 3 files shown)


transformers4rec/torch/masking.py
```diff
@@ -290,7 +294,8 @@ def _compute_masked_targets(
             label_seq_trg_eval[rows_ids, last_item_sessions] = labels[rows_ids, last_item_sessions]
             # Updating labels and mask
             labels = label_seq_trg_eval
-            mask_labels = label_seq_trg_eval != self.padding_idx
+            # We only mask padded positions
+            mask_labels = item_ids != self.padding_idx
```
gabrielspmoreira, line 298: Good catch!

gabrielspmoreira (approved):
Great job @sararb

PR #802, opened 2025-06-24
Sec pic fix

jperez999:
Fixes the pickle security issue.


jperez999:
ran all tests in tests/ in local environment and they all passed.

PR #816, opened 2026-04-17
Fix TypeError in Model.load on Python 3.9+

### LINKED ISSUE #810: [BUG] `Model.load` raises TypeError on Python 3.9+ (parameterized generic in isinstance)
### Bug description

After PR #809 (commit `8bf122f5`), `transformers4rec.torch.model.base.Model.load()` uses a parameterized generic as the second argument to `isinstance`:

https://github.com/NVIDIA-Merlin/Transformers4Rec/blob/8bf122f5dcb39feecfc6dabde734d79c2d1c4380/transformers4rec/torch/model/base.py#L917

```python
if isinstance(state_dict, Dict[str, torch.Tensor]):
    model.load_state_dict(state_dict, strict=strict)
else:
    raise ValueError("`state_dict` must be a dictionary of parameter (torch) tensors.")
```

On Python 3.9+ this raises `TypeError: Subscripted generics cannot be used with class and instance checks` before any of the load logic runs. Net effect: `Model.load()` is completely unusable on current `main`.

The regression is not caught by the existing test `tests/unit/torch/model/test_model.py::test_save_next_item_prediction_model`, which is the only place in the test suite that reaches this line. CI does not appear to exercise it on a clean env (likely the `copy-pr-bot` runners are currently blocked, see #798).

Pure-Python repro (no T4Rec install needed):
```python
from typing import Dict
import torch

isinstance({"a": torch.zeros(1)}, Dict[str, torch.Tensor])
# TypeError: Subscripted generics cannot be used with class and instance checks
```

Library-level repro:
```python
import torch
from transformers4rec.torch.model.base import Model

Model.load({"a": torch.zeros(1)}, heads=[])   # raises TypeError before anything else
```

### Expected behavior

`Model.load()` should accept a plain `dict` (as returned by `state_dict()` or `torch.load`) and load the weights into the provided heads.

### Environment details

- Transformers4Rec: `main` @ `8bf122f5`
- Python: any 3.9+ (documented semantics; all modern installs affected)
- PyTorch: any

### Additional context

Minimal fix — replace the parameterized generic with a plain `dict`:

```python
if isinstance(state_dict, dict):
    model.load_state_dict(state_dict, strict=strict)
else:
    raise TypeError("`state_dict` must be a dict of torch.Tensor.")
```

The tighter "dict of `str` → `torch.Tensor`" check cannot be done in a single `isinstance` call; if such validation is desired it has to be a loop over items. In practice `load_state_dict` itself will surface incompatible entries, so the bare `dict` check is sufficient.

Happy to send a one-line PR.

shaun0927:
### Goals :soccer:

Fix the `TypeError` raised by `Model.load(...)` on every Python >= 3.9 install.

Fixes #810.

After PR #809 (`8bf122f5`), `transformers4rec/torch/model/base.py:917` uses `isinstance(state_dict, Dict[str, torch.Tensor])`. `typing.Dict[str, torch.Tensor]` is a parameterized generic — passing it as the second argument to `isinstance` raises `TypeError: Subscripted generics cannot be used with class and instance checks` on Python >= 3.9 (per the CPython docs). The exception fires before any load logic runs, so `Model.load(...)` is unreachable on all current Python interpreters.

Minimal fix — narrow the type check to the concrete `dict` class:

```diff
-if isinstance(state_dict, Dict[str, torch.Tensor]):
+if isinstance(state_dict, dict):
     model.load_state_dict(state_dict, strict=strict)
 else:
     raise ValueError("`state_dict` must be a dictionary of parameter (torch) tensors.")
```

The `load_state_dict` call on the next line already surfaces structural problems with the dict contents (wrong keys, wrong tensor shapes, non-tensor values), so the narrower `Dict[str, torch.Tensor]` check never provided additional safety — it just failed earlier with a confusing error. The `raise ValueError(...)` branch and message are left untouched, matching the choice made in PR #808/#809.

### Testing Details :mag:

Pre-fix repro (Python 3.12):
```python
>>> from typing import Dict
>>> import torch
>>> isinstance({"a": torch.zeros(1)}, Dict[str, torch.Tensor])
TypeError: Subscripted generics cannot be used with class and instance checks
```

Post-fix behavior:
```python
>>> from transformers4rec.torch.model.base import Model
>>> Model.load({"a": torch.zeros(1)}, heads=some_head)   # no TypeError; proceeds to load_state_dict
>>> Model.load("/some/path", heads=some_head)
ValueError: `state_dict` must be a dictionary of parameter (torch) tensors.
```

The existing `tests/unit/torch/model/test_model.py::test_save_next_item_prediction_model` exercises this exact code path via `tr.Model.load(state_dict, new_model.heads)`. No test change is needed because the test previously could never have passed on Python >= 3.9.

### DIFF
--- a/transformers4rec/torch/model/base.py
+++ b/transformers4rec/torch/model/base.py
@@ -914,7 +914,7 @@ class Model(torch.nn.Module, LossMixin, MetricsMixin):
             max_sequence_length=max_sequence_length,
             top_k=top_k,
         )
-        if isinstance(state_dict, Dict[str, torch.Tensor]):
+        if isinstance(state_dict, dict):
             model.load_state_dict(state_dict, strict=strict)
         else:
             raise ValueError("`state_dict` must be a dictionary of parameter (torch) tensors.")


shaun0927:

shaun0927:
Amended to narrow the diff to `+1 -1`. The earlier commit also switched the exception type from `ValueError` to `TypeError`, which is out of scope for fixing the `TypeError: Subscripted generics...` bug and was inconsistent with the `raise ValueError` choice in PRs #808 and #809. Restored to the original `ValueError`.

Only commit now: `isinstance(state_dict, Dict[str, torch.Tensor])` → `isinstance(state_dict, dict)`.

PR #817, opened 2026-04-17
Fix TypeError in Head.build when body lacks output_size

### LINKED ISSUE #811: [BUG] `Head.build` raises `TypeError: 'NoneType' object is not callable` when body lacks `output_size`
### Bug description

PR #802 (commit `ab7207cf`, "Sec pic fix") changed the default sentinel in `transformers4rec/torch/model/base.py` from `lambda: None` to `None`:

https://github.com/NVIDIA-Merlin/Transformers4Rec/blob/8bf122f5dcb39feecfc6dabde734d79c2d1c4380/transformers4rec/torch/model/base.py#L288

```python
if not getattr(self.body, "output_size", None)():
    raise ValueError(
        "Can't infer output-size of the body, please provide  "
        "a `Block` with a output-size. You can wrap any torch.Module in a Block."
    )
```

When `self.body` does not have an `output_size` attribute, `getattr(..., None)` returns `None`, and the trailing `()` then evaluates `None()`, which raises `TypeError: 'NoneType' object is not callable`.

The original `lambda: None` sentinel was callable and returned a falsy value, so the intended `raise ValueError(...)` path below was actually reached. The current code path short-circuits with the wrong exception type, making the error message unactionable.

This is not caught by tests because all fixtures provide bodies with `output_size`.

```python
class Body:
    pass

body = Body()
getattr(body, "output_size", None)()
# TypeError: 'NoneType' object is not callable
```

Versus the pre-#802 behavior:
```python
getattr(body, "output_size", lambda: None)()
# None   (falsy, falls through to `raise ValueError(...)` in Head.build)
```

### Expected behavior

When the body lacks `output_size`, `Head.build` should raise the original `ValueError("Can't infer output-size of the body ...")` so the user gets a clear, actionable message.

### Environment details

- Transformers4Rec: `main` @ `8bf122f5` (regression introduced by PR #802 / commit `ab7207cf`)
- Python: any

### Additional context

Two equivalent minimal fixes:

Option A — restore the callable sentinel:
```python
if not getattr(self.body, "output_size", lambda: None)():
```

Option B — explicit two-step check (slightly more readable):
```python
output_size_fn = getattr(self.body, "output_size", None)
if output_size_fn is None or not output_size_fn():
    raise ValueError(...)
```

Option A preserves behavior byte-for-byte. Happy to send a PR.

shaun0927:
### Goals :soccer:

Restore the descriptive `ValueError("Can't infer output-size of the body ...")` that was inadvertently shadowed in PR #802 by a `TypeError: 'NoneType' object is not callable`.

Fixes #811.

`transformers4rec/torch/model/base.py:288` uses a callable sentinel so that a body without an `output_size` method falls through to the `raise ValueError(...)` branch:

```python
if not getattr(self.body, "output_size", None)():
    raise ValueError(
        "Can't infer output-size of the body, please provide  "
        "a `Block` with a output-size. You can wrap any torch.Module in a Block."
    )
```

The previous sentinel was `lambda: None` (a callable that returns a falsy value). PR #802 (`ab7207cf`, "Sec pic fix") replaced the sentinel with `None`. `None()` is not callable — it raises `TypeError` *before* the `not` is evaluated — so users who hit this path get a confusing `'NoneType' object is not callable` instead of the documented "Can't infer output-size of the body" guidance.

Minimal fix — restore the callable sentinel:

```diff
-if not getattr(self.body, "output_size", None)():
+if not getattr(self.body, "output_size", lambda: None)():
```

The semantics are byte-for-byte identical to the pre-#802 code. This is the smallest change that preserves the original intent without touching PR #802's other security improvements.

### Testing Details :mag:

Pre-fix behavior (reproduces the regression):
```python
class Body: ...
getattr(Body(), "output_size", None)()
# TypeError: 'NoneType' object is not callable
```

Post-fix behavior:
```python
getattr(Body(), "output_size", lambda: None)()
# None   (falsy; falls through to raise ValueError(...) in Head.build)
```

No new unit test is added: every existing body fixture in `tests/` provides `output_size`, so there is no negative case to regression-test without introducing a new stub. Happy to add one if reviewers prefer — a `torch.nn.Module`-only body (without the `output_size` method that `Block` exposes) is the natural shape.

### DIFF
--- a/transformers4rec/torch/model/base.py
+++ b/transformers4rec/torch/model/base.py
@@ -285,7 +285,7 @@ class Head(torch.nn.Module, LossMixin, MetricsMixin):
         device
         task_blocks
         """
-        if not getattr(self.body, "output_size", None)():
+        if not getattr(self.body, "output_size", lambda: None)():
             raise ValueError(
                 "Can't infer output-size of the body, please provide  "
                 "a `Block` with a output-size. You can wrap any torch.Module in a Block."


shaun0927:

